PR #15, opened 2020-05-06
Update CONTRIBUTING.md

finnlewis:
Suggesting that we include the full content of the contrib guidelines from https://docs.google.com/document/d/1CJWkNMh6rjF6Ml-WwEwWgt9UPNpTEgODYJN3hCaySJk/edit#heading=h.h26x5w3wy55f

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,76 @@
-# Contribution guidelines
+# LocalGov Drupal: Development and contribution processes and standards
+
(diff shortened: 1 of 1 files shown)


stephen-cox:
I do think there's quite a bit there which is not entirely relevant to a developer just trying to understand what's required of them. I can see myself skimming it, getting bored and moving on before getting to the stuff that I need to take notice of. But happy to merge this and discuss later on.

Issue #24, opened 2020-05-06
Add a default settings.php deployed with composer


stephen-cox:
The default install doesn't add a settings.php and the default doesn't have the code to include a settings.local.php enabled by default (it's commented out). For Lando or to run automated tests we want a default settings.php setup so we can easily add a settings.local.php for the appropriate environment.

Drupal's composer scaffold should make this easy to configure: https://www.drupal.org/docs/develop/using-composer/using-drupals-composer-scaffold

stephen-cox:
Closed with https://github.com/localgovdrupal/localgov/pull/26

PR #26, opened 2020-05-06
Document running tests locally

stephen-cox:


stephen-cox:
@finnlewis @andybroomfield I have added a few tests to the LocalGov profile and updated the docs for running tests in Lando. I still need to document getting Lando fully setup so people don't need to mess around with settings files. I'll do this as part of https://github.com/localgovdrupal/localgov/issues/24, but interested in getting your feedback on the approach so far.

Issue #43, opened 2020-05-20
Tests failing due to broken schema in contrib modules


stephen-cox:
Tests fail once the localgov_paragraphs module is enabled due to broken schema in the Geolocation and Office Hours modules; a SchemaIncompleteException is thrown. There's a patch for Office Hours, but not for Geolocation.

One solution is to add `$strictConfigSchema = FALSE` to the tests to disable schema checking. But this then fails code checks with: "Do not disable strict config schema checking in tests. Instead ensure your module properly declares its schema for configurations." So to use this we will need to remove the following check from our phpcs.xml.dist file: DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php

finnlewis:
Just confirming the behaviour locally before trying to fix. 

$ lando phpunit
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

Testing 
Drupal\Tests\localgov\Functional\LocalGovProfile
 ✘ Local gov drupal profile
   │
[...]

PR #44, opened 2020-05-20
Disable schema checking in tests due to broken schema in contrib modules

stephen-cox:
See https://github.com/localgovdrupal/localgov/issues/43

### DIFF
--- a/tests/src/Functional/LocalGovProfileTest.php
+++ b/tests/src/Functional/LocalGovProfileTest.php
@@ -17,6 +17,15 @@ class LocalGovProfileTest extends BrowserTestBase {
   protected $profile = 'localgov';
 
   /**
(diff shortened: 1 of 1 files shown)


stephen-cox:
Tests include a schema check, but these fail due to contrib modules not including correct schema. See #43, for a further discussion.
Any thoughts @finnlewis @ekes. There maybe a better way to do this.

finnlewis:
I think we want to keep the schema checking if we can. @ekes sounded keen to keep it. 
I've done some work on the geolocation and office_hours patches on https://github.com/localgovdrupal/localgov/issues/43
If that all works, I guess we knock this one back?

stephen-cox:
Yep - rejecting this pull request as we've found a better solution.

Issue #49, opened 2020-05-22
Sitemap module


finnlewis:
Include sitemap module
See https://github.com/localgovdrupal/localgov_services/issues/11

finnlewis:
I've had issues getting the configuration for simple_sitemap module to install in the desired manner. 

See related pull requests:
https://github.com/localgovdrupal/localgov/pull/50 
https://github.com/localgovdrupal/localgov_services/pull/53

We would ideally like the config to inclde each content type in the sitemap to be optional, depending on the simple_sitemap modile being installed. 

This is how it is currently trying to work, for example on the localgov_guide content types: 
https://github.com/localgovdrupal/localgov_guides/blob/master/config/optional/simple_sitemap.bundle_settings.default.node.localgov_guides_overview.yml

Basically, this is not importing.

Interestingly it does import if the config is in the /config/install directory, just not if it is in the /config/optional folder. 

I've posted an issue on drupal.org: https://www.drupal.org/project/simple_sitemap/issues/3156080

ekes:
If it works in config/install when simple_sitemap isn't installed? But what about install order?

My thoughts were:
 * Move to profile, but still have to handle optional modules enabled later.
Or maybe better:
 * `\Drupal::service('simple_sitemap.generator')->setBundleSettings('node', $content_type, ['index' => TRUE, 'priority' => 0.5])` in install and modules_installed hooks?

PR #50, opened 2020-05-22
Feature/49 sitemap

finnlewis:


ekes:
Where do we want to put enabling it, and any shared configuration?

The entity / type specific stuff should live as optional configuration I think. But are there global settings?

finnlewis:
I guess enabling it in https://github.com/localgovdrupal/localgov/blob/master/localgov.info.yml makes sense for now... then figure out what, if any specific config needs initiating?

ekes:
Is there likely to be a lot more straight SEO stuff. Such that it would want its own submodule?

That said unless it's obvious. Can put in the .info.yml for now; and move if it makes sense.

finnlewis:
How's that @ekes ?
I guess that config for each content type needs to live with that module, but at least it is now there and enabled by default.

ekes:
> I guess that config for each content type needs to live with that module, 

Yes. I'd put the services ones in optional:

https://github.com/localgovdrupal/localgov_services/blob/master/modules/localgov_services_landing/config/optional/simple_sitemap.bundle_settings.default.node.localgov_services_landing.yml

ekes:
For the rest. Having it in the profile level probably makes sense. There's nothing else that it does. And if there is more SEO it can be moved.

finnlewis:
@ekes I've now added simple_sitemap and some default configuration.

Also moved the existing optional configuration to default on https://github.com/localgovdrupal/localgov_services/pull/53

finnlewis:
Hi @ekes 

I've updated the issue with my findings here https://github.com/localgovdrupal/localgov/issues/49

Since the optional configuration does not appear to be working, a good starting point is just to include the configuration for simpel_sitemap in the profile for the services content types. 

We can address the others (guides / step by step) in their own modules.

How does that sound for now?

Cheers,

Finn

finnlewis:
@stephen-cox the reason it is here and not in the services module, is that the services modules might be installed without the profile and without simple_sitemap, and the config/optional does not work for the simple_sitemap bundle config. 

We can move later if we resolve that. For now, at least we have sensible defaults on install.

PR #51, opened 2020-05-25
Feature/33 services pages

ekes:
The basic functionality of Services (the three content types) their integration linking, and related dependencies paragraphs, page components, topics, in services master should all be progressed enough to make it stable enough to start integrating.

(Suggest a squashed merge - as it's really just one line in the end - the review is the master branch of the module/s themselves!)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,10 +8,6 @@
         {
             "type": "composer",
             "url": "https://packages.drupal.org/8"
-        },
-        {
-            "type": "vcs",
-            "url":  "git@github.com:localgovdrupal/localgov_services.git"
         }
     ],
     "require": {
@@ -21,8 +17,8 @@
         "drupal/devel": "^2.1",
         "drush/drush": "~9",
         "localgovdrupal/localgov_core": "dev-master",
-        "localgovdrupal/localgov_services": "dev-develop",
-        "localgovdrupal/localgov_paragraphs": "dev-feature/32-core-paragraphs",
+        "localgovdrupal/localgov_services": "dev-master",
+        "localgovdrupal/localgov_paragraphs": "dev-master",
         "localgovdrupal/localgov_theme": "dev-master"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


finnlewis:
I'm just testing this branch by changing the composer.json in the project root to use 

 `"localgovdrupal/localgov": "dev-feature/33-services-pages"`

Then running 

`composer updated --no-cache`

Followed by

finnlewis:
Dependency issues when enabling localgov_services_landing
`
Drupal\Core\Extension\MissingDependencyException: Unable to install modules: module 'localgov_services_landing' is missing its dependency module inline_entity_form. in Drupal\Core\Extension\ModuleInstaller->install() (line 114 of core/lib/Drupal/Core/Extension/ModuleInstaller.php). `

ekes:
> Dependency issues when enabling localgov_services_landing
> `Drupal\Core\Extension\MissingDependencyException: Unable to install modules: module 'localgov_services_landing' is missing its dependency module inline_entity_form. in Drupal\Core\Extension\ModuleInstaller->install() (line 114 of core/lib/Drupal/Core/Extension/ModuleInstaller.php).`

Hmm... I'd think composer should be pulling it https://github.com/localgovdrupal/localgov_page_components/blob/master/composer.json#L10 No?

finnlewis:
I now have the installation working and providing content types.
However, creating any node seems to give this error: 
`Twig\Error\SyntaxError: Unknown "getStatusMessageWrapperClass" function. in Twig\ExpressionParser->getFunctionNodeClass() (line 30 of themes/contrib/localgov_theme/templates/misc/status-messages.html.twig). `

finnlewis:
That twig error is related to the theme and dependency on the getStatusMessageWrapperClass function which used to be defined in the bhcc_helper module.

ekes:
Hmm.. I've been testing this without the theme (just bartik and claro); and the tests use stark iirc.

Looking at the theme:
https://github.com/localgovdrupal/localgov_theme/blob/master/templates/misc/status-messages.html.twig#L30

It has dependencies on bhcc_helper and we're not importing that I think. So Should the theme be on yet?

PR #62, opened 2020-06-04
Place some default blocks on site install

stephen-cox:
This places a number of system and related blocks on a site install.

There are a number of related merge requests:

- https://github.com/localgovdrupal/localgov_theme/pull/22
- https://github.com/localgovdrupal/localgov_core/pull/10

To test this you will need to checkout the https://github.com/localgovdrupal/localgov_project/tree/feature/60-default-blocks branch in your project repo and run `composer update --no-cache`

If a composer update doesn't pull down the related branches try the nuclear option:

```
rm -fr web/profiles/
rm -fr web/modules/
rm -fr web/themes/
composer cc
composer update --no-cache
```

Once we're happy with this merge request please don't merge it. I need to change the `composer.json` file before it can be merged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,10 +16,10 @@
         "drupal/console": "~1.0",
         "drupal/devel": "^2.1",
         "drush/drush": "~9",
-        "localgovdrupal/localgov_core": "dev-feature/14-move-block-config",
+        "localgovdrupal/localgov_core": "dev-master",
         "localgovdrupal/localgov_services": "dev-master",
         "localgovdrupal/localgov_paragraphs": "dev-master",
-        "localgovdrupal/localgov_theme": "dev-feature/14-header-and-footer"
+        "localgovdrupal/localgov_theme": "dev-master"
     },
     "require-dev": {
         "drupal/core-dev": "^8.8"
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -16,10 +16,10 @@
         "drupal/console": "~1.0",
         "drupal/devel": "^2.1",
         "drush/drush": "~9",
-        "localgovdrupal/localgov_core": "dev-master",
+        "localgovdrupal/localgov_core": "dev-feature/14-move-block-config",
```
ekes, line 19: Want to remember _not_ to merge this, but to merge all other branches at the same item.

finnlewis:
I agree @ekes that having a super custom node title block feels possibly too opinionated, but it is the sort of thing we can swap out easily enough later. In the interests of having something that starts to look like a site, I'm OK with it for now.

stephen-cox:
> The only thing I'm now a tiny bit concerned about is the node title thing.

Agreed that the node title block is not ideal. There's also a taxonomy title block to port and then all views and non-entity stuff also need to define their own titles, but as Finn says we can deal with this in due course. One option is to extend the system title block to handle the lede stuff.

stephen-cox:
Thanks all. Have undone the composer.json changes used for testing. Now merging.

PR #115, opened 2020-09-03
Patch to fix for nested Paragraphs create multiple drag handles https…

finnlewis:
…://www.drupal.org/project/drupal/issues/3092181

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,5 +37,15 @@
         "sort-packages": true
     },
     "extra": {
+      "enable-patching": true,
+      "composer-exit-on-patch-failure": true,
+      "patchLevel": {
+        "drupal/core": "-p2"
+      },
+      "patches": {
+        "drupal/core": {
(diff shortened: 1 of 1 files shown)


finnlewis:
@Adnan-cds @stephen-cox - I've tested this patch and it does seem to fix the multiple drag handles on nested paragraphs. 
The tests all pass, so I think this should be OK to merge.

composer.json
```diff
@@ -37,5 +37,15 @@
         "sort-packages": true
     },
     "extra": {
+      "enable-patching": true,
```
Adnan-cds, line 40: As far as I know, "enable-patching" is only relevant for [package level](https://github.com/cweagans/composer-patches#allowing-patches-to-be-applied-from-dependencies) (AKA root) composer.json files.  [localgovdrupal/localgov_project](https://github.com/localgovdrupal/localgov_project/blob/65742666e3699bbda6fd72ede9911f32ffb72a22/composer.json#L38) already has ```enable-patching``` set to ```true``` making this a bit redundant :(

Adnan-cds (approved):
Happy for this patch to go in.

PR #137, opened 2020-09-24
Update README to modify creatre-project to bypass possible memory_limit

tanc:
Add COMPOSER_MEMORY_LIMIT=-1 to the create-project command

This is to ensure the host php memory_limit doesn't stop the install from running out of memory as is possible in some situations.


stephen-cox (approved):
Looks good :+1:

PR #143, opened 2020-10-06
Upgrade to Drush 10.

ekes:
Recommended version for Drupal 8.4+ https://www.drush.org/install/#drupal-compatibility So will also be needed for 9. 
Supports the new configuration ExportStorage, which means new config management tools will use it, and so will https://github.com/localgovdrupal/localgov_directories/blob/master/src/EventSubscriber/DirectoriesConfigSubscriber.php

You will need to `drush cr` and `drush cc drush`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "drupal/console": "~1.0",
         "drupal/devel": "^2.1",
         "drupal/simple_sitemap": "^3.7",
-        "drush/drush": "~10",
+        "drush/drush": ">=10",
         "localgovdrupal/localgov_alert_banner": "dev-master",
         "localgovdrupal/localgov_core": "dev-master",
         "localgovdrupal/localgov_campaigns": "dev-master",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "drupal/console": "~1.0",
         "drupal/devel": "^2.1",
         "drupal/simple_sitemap": "^3.7",
-        "drush/drush": "~9",
+        "drush/drush": "~10",
         "localgovdrupal/localgov_alert_banner": "dev-master",
         "localgovdrupal/localgov_core": "dev-master",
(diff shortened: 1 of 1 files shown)


Adnan-cds:
Hi Ekes,
I am fine with Drush 10.  I had no problem with Drush 10 while testing localgov_directories on a stock Drupal site recently.  So we should be okay.  I just find it a bit odd that Drush has been added to the **localgov profile** rather than the project template (i.e. localgovdrupal/localgov_project).  When Drush 11 will be ready for use with LocalGov, we will have to update the profile again.  If it were in the project template, people would just update their own template and move on.  Let me know if you want Drush to remain in the profile or not.

ekes:
Looking through profiles of lightning, thunder I see it in lightning not thunder. But what I notice is lightning it's a dev requirement, and it's >= which wouldn't stop anyone putting a higher version in their top composer.json even if it wasn't dev either. 

That would be a way of suggesting the profile really does need a minimum version.

Otherwise the project does seem to make sense.

Adnan-cds:
> That would be a way of suggesting the profile really does need a minimum version.

That makes sense.  You may have good reasons to ask for a minimum Drush version.

ekes:
Should probably merge this separately; and then make another issue about if it lives in require-dev and/or the localgov-project composer.json.

stephen-cox:
I think I added this to the profile because Drush seems essential for all Drupal sites and that was what some other distributions have done. Dev dependencies in second-level composer files are not installed, but I don't know whether the version numbers are respected.

stephen-cox (approved):

Adnan-cds:
> and then make another issue about if it lives in require-dev and/or the localgov-project composer.json.

Now that you have switched to a more generous minimum version constraint, I am happy for drush/drush to be in the *require* section rather than *require-dev*

Adnan-cds (approved):
Looks good :)

PR #155, opened 2020-11-06
Update README.md

finnlewis:


finnlewis:
@stephen-cox @Adnan-cds @andybroomfield 

My assumption is that these sort of documentation updates don't necessarily warrant peer review, but what do you think? Maybe they should for legibility and spelling... (goes to check the spelling again)

README.md
```diff
@@ -1,19 +1,45 @@
 # LocalGov Drupal
 
 Drupal distribution and install profile to help UK councils collaborate and
-share Drupal code.
+share Drupal code for publishing website content.
 
-## Installing LocalGov Drupal
+## Requirements for instaling LocalGov Drupal locally
```
Adnan-cds, line 6: s/instaling/installing/ :)

Adnan-cds:
> Maybe they should for legibility

That's right, the review should help with legibility.  But I am fine if someone else from the **same** organization approves it before it is merged.

andybroomfield:
Yeah, ok for these to merge, just have someone proof read it.

We might want to have more peer review when we document how to extend localgov and the api the custom modules have.

README.md
Adnan-cds: If you please do something about this.

README.md
```diff
@@ -33,7 +59,7 @@ seems to be a solid approach:
 
 ```bash
 
-rm -rf web/profiles/contrib/ web/modules/contrib/ web/themes/contrib/;
+rm -rf web/;
```
Adnan-cds, line 62: Hi Finn,
Any chance you can get a second opinion on this from Steve who I wrote it first?  I don't know why but I am feeling a bit uncomfortable wiping out the entire "web/" where we may have custom modules, settings files, etc. :(

andybroomfield: Second on this, it seems odd to say rm the whole web directory.

finnlewis: Agreed - it's a bit of an assumption. It does work if you're just installing it all from composer for dev purposes and blowing it away repeatedly, but I agree this is not an assumption we necessarily want to make.

PR #157, opened 2020-11-10
Reaplace localgov_subsites with localgov_campaigns.

finnlewis:


finnlewis:
@Adnan-cds @andybroomfield 

Let us know how this is working for you and your migrations.

composer.json
```diff
@@ -27,7 +26,8 @@
         "localgovdrupal/localgov_paragraphs": "^1.0@alpha",
         "localgovdrupal/localgov_services": "^1.0@alpha",
         "localgovdrupal/localgov_step_by_step": "^1.0@alpha",
-        "localgovdrupal/localgov_theme": "^1.0@alpha"
+        "localgovdrupal/localgov_subsites": "^1.0@alpha",
+	"localgovdrupal/localgov_theme": "^1.0@alpha"
```
Adnan-cds, line 30: The indentation is not looking right :(

Adnan-cds:
I can confirm that you **cannot** keep both the localgov_campaigns and localgov_subsites modules **active** at the same time.  This is because the localgov_subsites:localgov_subsites_paragraphs submodule declares the same Paragraphs as localgov_campaigns:localgov_campaigns_paragraphs.  So Drupal just refuses to enable localgov_subsites_paragraphs.

This means the only way to enable the localgov_subsites module is to delete all Paragraph entities whose Paragraph type is defined in localgov_campaigns_paragraphs followed by uninstalling localgov_campaigns_paragraphs.  But once you have deleted these Paragraphs entities, you cannot use them from the subsites related content types.  So I don't see a migration path.  But this shouldn't be a problem for new completely new sites.

For the Croydon site, I will next try to:
- Rollback the Paragraph migrations.
- Uninstall localgov_campaigns and localgov_campaigns_paragraphs modules.
- Install the localgov_subsites module.
- Create necessary migration script for subsites related content types and then rerun the migration.

I will keep you posted about the migration.

But please note that if anyone has a LocalGov site with actual Campaign content, they will have difficulty swapping localgov_campaigns with localgov_subsites.

Adnan-cds:
I have not been able to enable the localgov_subsite_paragraphs submodule even after uninstalling the localgov_campaigns_paragraphs submodule.  This is because the Paragraph **configs** (e.g. localgov_call_out_box) from localgov_campaigns_paragraphs have already been imported into Drupal.  So when I try to install the localgov_subsites_paragraphs submodule, Drupal refuses to install these Paragraphs again.  This means right now the localgov_subsites_paragraphs submodule cannot be installed on an existing LocalGov site where localgov_campaigns has already been installed.  On solution could be to move all the config files of localgov_subsites_paragraphs from *config/install* into *config/optional*.

ekes:
How many of the paragraphs are subsite specific?

Are there better ways of packaging up paragraphs?

If we are already not making their configuration dependent on the supplying module (which admittedly could be an oversight I know we've forgotten it on other config - but 'call out box' sounds like I could see some one using them elsewhere in their site and then not realize that they were dependent on the module).

finnlewis:
@Adnan-cds paragraphs are becoming a bit of a burden.

Maybe we should move all the paragraphs to the localgov_paragraphs module, or even break them out more granularly, 

These decisions really expose the need for an overall information architecture design. 

Any thoughts @ekes @stephen-cox

ekes:
We exactly don't want to move them all into localgov_paragraphs module as the problem is then everyone gets all the paragraphs. Breaking them out into more module, with only the minimum required by other modules, the rest being optional seems the best way.

Adnan-cds:
> or even break them out more granularly,

+1

> Breaking them out into more module, with only the minimum required by other modules, the rest being optional seems the best way.

+1

Steve and I were [discussing similar](https://github.com/localgovdrupal/localgov_paragraphs/pull/6#issuecomment-682462352) while I was working on the Homepage-specific paragraphs.  Back then we added a localgov_homepage_paragraphs submodule to localgov_paragraphs.  We can take this further and add more submodules.  People can then selectively enable these submodules.


FAO @stephen-cox

iam-jam:
Folks I've added a commit to this branch that add a composer alias to a fork of the Drupal dbal module. DBAL is required by entity_hierarchy which is a dependency of localgov_subsites. The forked version of DBAL includes a fix that allows us to use the module with SQLite without causing notices and consequently causing PHPUnit tests to fail. See the following for more info: 

- https://www.drupal.org/project/dbal/issues/3185995

Adnan-cds:
> I've added a commit to this branch that add a composer alias

I think this needs to go into the [localgov_project's composer.json](https://github.com/localgovdrupal/localgov_project/blob/master/composer.json) file.  You only grab this file once at the beginning of your LocalGov site.  So even with a change there, only new sites will benefit from this change.

It hit me only now that while repositories can only be declared in a root-level composer.json file, patch files can be added anywhere.  So one alternative for you would be to add a patch file to the d.o ticket and then refer to it from localgov_subsite's composer.json file.  It's a bit low tech but should do the job :)

iam-jam:
Thanks @Adnan-cds I'll go with the patch then :)

finnlewis:
So, now that we've got the tests passing on localgov_subsites, are we ready to merge this into the default codebase install?

composer.json
Adnan-cds: Hi Finn, this indentation is still not looking right :(

FAO @finnlewis

Adnan-cds:
There are some indentation issues with this change, otherwise it looks okay now.

finnlewis:
@Adnan-cds indentation resolved, tests pass, OK to merge?

Adnan-cds (approved):
All looks good now.  Thanks a lot :)

PR #170, opened 2021-01-05
Patch Drupal core to avoid tests breaking in PHP 7.4.

finnlewis:


finnlewis:
@stephen-cox this PR should help with the tests in PHP 7.4, but it is still testing against PHP 8.

If we can try this with the PHP 7.4 docker image, I am hoping this will get the tests running better again.

CC @Adnan-cds  @andybroomfield  @ekes

finnlewis:
@Adnan-cds @andybroomfield  @ekes @stephen-cox 
This is one of those cases where the tests are not applying the composer.json patch as it is not using it;s own self branch to install via composer. 

Patch is applying and tests are passing locally.

Could one of you just give this a quick look and approve the merge?

I think this will unblock the other testings.

Adnan-cds:
> This is one of those cases where the tests are not applying the composer.json patch as it is not using it;s own self branch to install  via composer.

Would you mind trying the [template test runner ](https://github.com/localgovdrupal/.github/blob/master/workflow-templates/localgov-drupal-ci.yml) for [this repo](https://github.com/localgovdrupal/localgov/blob/master/.github/workflows/test.yml).

Adnan-cds:
The [d.o patch](https://www.drupal.org/files/issues/2021-01-05/fix-php74-tests-3191012-2.patch) looks harmless to me.  Thanks :)

Adnan-cds (approved):
Looks good to me.  Thanks :)

finnlewis:
Yay! Good call @Adnan-cds ! :+1:

PR #212, opened 2021-05-07
Drupal 9 support

### LINKED ISSUE #182: Drupal 9 test failing: Local Gov Profile (Drupal\Tests\localgov\Functional\LocalGovProfile
```
Testing 
Local Gov Profile (Drupal\Tests\localgov\Functional\LocalGovProfile)
 ✘ Local gov drupal profile
   ┐
   ├ TypeError: Argument 2 passed to PHPUnit\Framework\Assert::assertContains() must be iterable, string given, called in /app/web/profiles/contrib/localgov/tests/src/Functional/LocalGovProfileTest.php on line 31
   │
   ╵ /app/vendor/phpunit/phpunit/src/Framework/Assert.php:169
   ╵ /app/web/profiles/contrib/localgov/tests/src/Functional/LocalGovProfileTest.php:31
   ╵ /app/vendor/phpunit/phpunit/src/Framework/TestResult.php:722
   ┴

### PR DESCRIPTION
**Pull request by stephen-cox**
Fixes #182 

Part of #197

### DIFF
--- a/localgov.info.yml
+++ b/localgov.info.yml
@@ -16,7 +16,6 @@ install:
   - drupal:dblog
   - drupal:dynamic_page_cache
   - drupal:editor
-  - drupal:entity_reference
   - drupal:field_ui
   - drupal:file
   - drupal:help

### CODE REVIEW

File: composer.json [hunk 1]
```diff
@@ -12,8 +12,8 @@
     ],
     "require": {
         "drupal/admin_toolbar": "^2.2",
-        "drupal/core": "^8.9",
-        "drupal/devel": "^2.1",
+        "drupal/core": "^8.9||^9.1",
```
ekes: I think we need to be careful what we are promising here.

^9.1 is everything <10.0 So within this major version 1.* we're saying this profile is going to support Drupal 9.9 With our own code we might choose to do this, but with dependencies it's going to be more challenging. The point of bumping major versions is to make this easier.

Looking at Thunder and Lightning I notice they not only increase their major version with the major version of Drupal, but the minor version too. Thunder and Lightning are more tightly coupled that we are, and have more dependencies etc. so this is several steps removed from what we're proposing here, and felt to me initially like it's creating work, but I think I understand why:

- Dependencies. As anything is added that depend on new features of core the minimum version of core is going to have to increase. This will require increasing at least the minor version anyway. Doing this systematically makes could make this clearer. Module dependencies can do this too. So all of a sudden the minimum core version a dependency is requiring is increased, so the core version of the profile has to as well. It would cut this pitfall.


- Testing. Thunder and Lightning only have to test against one major.minor version of Drupal per major.minor version of their profile. If they support multiple versions, then they'd have to run multiple tests against the same profile. So that's got to make things simple.

For our dependencies we've already got the questionable situation of localgov events wanting to install two different major versions of a module to support 8 and 9. Even if development is slow there and it doesn't break now it seems to be a dangerous precedent and pattern to follow. 

So while I don't think we need to follow Thunder and Lightning's example, we will have to prepare for the point of where features in a specific higher version of core, or dependencies are requiring a specific version. This means being able to have supported releases parallel.

This contract should be clear to new adopters of the profile. [Unless we're dropping SemVer and going to go rolling releases and support Drupal 9 onwards in 1.0.0]. Don't install 1.0.0 it's not where development and support for Drupal 9 is going to go forward. It should equally be simple for existing users to understand.

In summary 
 - if we want 1.0.0 to support Drupal 9, we shouldn't be suggesting it is going to support 9.9. Fix it at a version we know;
 - unless we're dropping semver we're going to have points in the future where we have two supported versions of the profile so we should prepare and have processes in place to be able to do this;
 - we're going to have to have a release structure that can progress along with Drupal 9 and with semver this is 2.0.0

ekes (changes requested):

composer.json
finnlewis: So is not simplest to release the profile at 1.0.0 supporting Drupal 8 only, then 2.0.x for Drupal 9?

Some modules at their 1.0.0 release will support both 8 and 9, others like localgov_events will probably have different major versions.

andybroomfield: Suggestion only:
[composer require drupal/core:"^8.9 | ~9.1"](https://semver.mwl.be/?package=drupal%2Fcore&constraint=%5E8.9%20%7C%20~9.1&stability=stable)

Looking at https://endoflife.date/drupal:
Drupal 9.1 will receive security coverage until December 8, 2021 when Drupal 9.3.0 is released.
Drupal 8.9.x as the LTS cycle for Drupal 8 will be supported until November, 2021.

So why not set version 1.0 of the profile to 8.9 | 9.1 only.
And version 2 as 9.1 upwards.
This allows time for people to transition to Drupal 9 with the current profile, and then upgrade the profile.

Or is that overcomplicated things?

stephen-cox: Agreed that we will need a 2.0.0 release that drops Drupal 8 support, the question in my mind is when and I think this can be decided by breaking this down into the pros and cons with the different approaches

If I've followed the conversation so far correctly, the two options we have are:

- Support Drupal 8.9 and 9.1 in the 1.x versions, create a 2.x version with 9.x support
- Support Drupal 8.9 in 1.x and 9.x in 2.x

Does that sound correct?

Initial thoughts on these approaches:

## Support Drupal 8.9 and 9.1 in the 1.x versions

### Pro

- Fixes and features are available for Drupal 8 and 9 sites without extra work backporting them to Drupal 8
- Creating a 2.x release can wait until it's required, which for some modules may not be until November.

### Con

- We'll need to add testing for Drupal 8 and 9 to 1.x versions
- We can't use Drupal 9 only features in the 1.x branch (although I'm not sure how much of problem this actually is).

### Support Drupal 8.9 in 1.x and 9.x in 2.x

### Pro

- Testing can continue largely as it is now as major versions are tied to a single major Drupal version, although I suspect we will need to make some changes (see my comment about branch naming below).

### Con

- New features and bug fixes need to be ported to different major versions.

## Other thoughts on this

There must be other pros and cons, sounds like a proper discussion on this would be useful.

I don't think having a single `master` branch is maintainable when we're supporting multiple versions. Should we rename `master` to `1.x` and create new `2.x` branches for Drupal 9?

Should the supported Drupal version be declared in each modules `composer.json`? This would help Composer select the correct module version, which is probably not necessary if 1.x versions only support Drupal 8.

andybroomfield: Thanks @stephen-cox thats useful to know.
[See my comment on Localgov_core](https://github.com/localgovdrupal/localgov_core/pull/68#issuecomment-835379447) which does have an issue with event dispatcher parameters
> I'm coming round to having explicit releases that are Drupal 9 only now (in a v2.0 branch), depending on if people feel this would simplify the maintenance, or more complicate things with maintaining both releases (I'd like to see the Drupal 8 v1.0 versions at least maintained and available till Drupal 8 end of life in November).

I think now for simplicitiy, we should have profiles that mirror Drupal core versions 1.0 means you install Drupal 8, and thats for councils that joined early so they can keep using it, and 2.0 which is Drupal 9 only and for all new site builds which the project template will change to install by default. There **must** be an upgrade path between 1.0 and 2.0 so when a council is ready to upgrade to Drupal 9, they can upgrade their profile and move over.

This will allow modules to go Drupal 9 only once they need to do something the Drupal 9 way, although the event dispatcher issue means for localgov_core that needs to go 2.0 now.

We need to decide the support level for the 1.0 versions. Will they be getting new features or will they be bug fix and security support only? Like the profile they **must** have an upgrade path so that when a site is ready to go Drupal 9 and that means installing 2.0 modules, that running the update.php script is enough to bring any data over.
And the support period for the 1.0 versions should be till Drupal 8 is EOL (or at least when we know all partner councils have made the upgrade).

andybroomfield:
Have the changes here been resolved @ekes @stephen-cox and this is ok for review / merging?

stephen-cox:
@andybroomfield This is still needed. As it's such a simple change I'll merge later today unless I hear otherwise

andybroomfield (approved):
I've just tried to install the d9 profile with the template and this branch checked out.
This is needed as Entity reference is now in core.

Issue #200, opened 2021-04-28
Make necessary changes to the composer.json files.


finnlewis:

ekes:
Highlight localgovdrupal/localgov_guides#46 as one that has a bit of work and a question before tagging.

ekes:
localgovdrupal/localgov_theme#178 Could probably do with some different eyes on it.

ekes:
When all above are closed then update composer.json here for ~1.0.0 version (changing from alpha modules and changing to patch only auto updates ^ to ~).

Then tag 1.0.0 for the profile itself.

stephen-cox:
Done - sprint 2

PR #214, opened 2021-05-10
Feature/210 tag 1.0.0

ekes:
Updates the composer.json to fit #200


danchamp (approved):

stephen-cox:
@ekes I assume it's intentional to version everything at `~1.0.0` - meaning patch updates will be installed but minor updates will need manual approval.

Edit: Should have read the issue first: https://github.com/localgovdrupal/localgov/issues/200#issuecomment-828560106

ekes:
re ~1.0.0 yes it was part of the discussion.

Basic logic: minor version updates are where it's likely that support for minor versions of dependencies will be dropped; this should go into a changelog and be notified clearly without automatically coming in. That's possible the more important, also making it explicit that new features are coming in is good.

PR #216, opened 2021-05-11
WIP: Github CI testing for 2.x

stephen-cox:


ekes:
```
    GIT_BASE: 2.x
    GIT_BRANCH: feature/github-ci-testing
    GIT_REPO: localgovdrupal/localgov
```

Looking good.

Was just wondering _if_ we wanted to run tests against feature/ bug/ branches and they are marked feature/2.x/123-description etc. could the pattern match. Or is that over-complicating as it's just on merge to the branches... or what about release: 2.x/release branch?

andybroomfield:
Will 2.0 profile still include modules on their 1.0 release? If so they will need to be tested against 1.0 and 20 (Drupal 8 and 9)

ekes:
> Will 2.0 profile still include modules on their 1.0 release? If so they will need to be tested against 1.0 and 20 (Drupal 8 and 9)

Yep. Good point.

stephen-cox:
I think for the `2.x` branches we can run just against Drupal 9 using the `2.x` version of the project. `1.x` branches are those that support D8 and D9 and so these will need to run against both these.

I'm trying to think how to generalise the CI stuff so we don't have to repeatedly update all repos. 

- Is it likely modules will go to a `3.x` release and still support D9 (i.e. move before D10 support is included and D9 support drop)? Just wondering if it's safe to assume all major module versions will support the same version of Drupal. I suspect it's not safe to assume this, but it will simplify the CI setup if we can.

stephen-cox:
> Was just wondering _if_ we wanted to run tests against feature/ bug/ branches and they are marked feature/2.x/123-description etc. could the pattern match. Or is that over-complicating as it's just on merge to the branches... or what about release: 2.x/release branch?

This is easy enough to do with setting the branches here: https://github.com/localgovdrupal/localgov/blob/615bf25da619f3c8f7a2694a99f2db7751b005a9/.github/workflows/test.yml#L6

Not sure about running tests whenever there's a push to a feature branch. Release branches should be tested with the PR, but could be included if there are other instances where we want to run tests.

ekes:
> Is it likely modules will go to a 3.x release and still support D9

Likely. Don't think I can run odds. The only case I could foresee already is if something like localgov_events gets developed further adds in full search_api support and it's a breaking change. Actually that's quite likely isn't it? Like moving events from straight sql views to swappable indexed backend. Might be large enough change for us to want to do that in a major rather than a minor. That would be 3.x

stephen-cox:
There's a dependency problem with having the profile and localgov_core moving to 2.x. There are modules that require the 1.x version of the localgov_core and these prevent Composer from installing the site; https://github.com/localgovdrupal/localgov_step_by_step/blob/d08068577aedb103fce299800dc08053823b01df/composer.json#L9

We will either have to release 2.x version of these modules or allow them to use 1.x or 2.x of core.

Thoughts?

stephen-cox:
Ended up creating a new PR for this #229 so closing

PR #228, opened 2021-05-17
Updated nested paragraphs patch for D9 support

### LINKED ISSUE #226: Nested paragraphs core patch fails on Drupal 9 install
```
  - Applying patches for drupal/core
    https://www.drupal.org/files/issues/2020-05-28/3092181-87-89x.patch (Fix for Nested Paragraphs create multiple drag handles https://www.drupal.org/project/drupal/issues/3092181)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2020-05-28/3092181-87-89x.patch
```

https://github.com/localgovdrupal/localgov/blob/9cd006dc0ddacf16d801ace29a9be7d606521b42/composer.json#L48

stephen-cox:
Use the latest D9 patch from: https://www.drupal.org/project/drupal/issues/3092181

Closes #226

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
       },
       "patches": {
           "drupal/core": {
-              "Fix for Nested Paragraphs create multiple drag handles https://www.drupal.org/project/drupal/issues/3092181": "https://www.drupal.org/files/issues/2020-12-08/3092181-142-9X_0.patch"
+              "Fix for Nested Paragraphs create multiple drag handles https://www.drupal.org/project/drupal/issues/3092181": "https://www.drupal.org/files/issues/2020-12-08/3092181-142-9X_0.patch",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -44,9 +44,7 @@
         "drupal/core": "-p2"
       },
       "patches": {
-        "drupal/core": {
```
finnlewis, line 47: Can I just check @stephen-cox , do we no longer need to specify what is being patched ( such as drupal/core ) when including patches like this?

stephen-cox:
@finnlewis I have fixed the error in the composer file. Tests won't run until #229 is merged

stephen-cox:
@finnlewis Pining you on this as I think you're the best place to check the patch is still doing what's needed and we should get it merged before doing a 2.0.0 release of the profile.

finnlewis:
@stephen-cox the original patch works for me. Had to add another patch for another bug in claro, which limited the width to 48rem on node/add forms.

Working better now!

finnlewis (approved):

stephen-cox:
Thanks @finnlewis - merging

Issue #225, opened 2021-05-13
Rename master branches to 1.x


stephen-cox:
To be able to easily support multiple major module versions we need to rename current master branches to 1.x. We'll then create new 2.x branches with breaking changes, as per the [versioning guidelines](https://github.com/localgovdrupal/localgov/wiki/Versioning).

We'll need to rename the master branch, fix the GitHub Actions as these are tied to master, then make a 1.0.x release with the changes.

We may want to merge the new 1.x into any open PR branches to ensure the GitHub CI stuff works for them·

PR #229, opened 2021-05-18
Updated Github CI to run against 2.x branch

stephen-cox:
Part of #225


finnlewis (approved):
Tests pass, looks good to me, so we're rattling through these to get them all in.

finnlewis (approved):
Tests pass, looks good to me, so we're rattling through these to get them all in.

PR #237, opened 2021-05-19
Update requirements for 2.0.0 versions for a stable 2.0.0 release for…

finnlewis:
… Drupal 9 support.

PR #209, opened 2021-05-04
Test database updates

stephen-cox:
Closes #196

PR #238, opened 2021-05-19
2.0.0 release

stephen-cox:
Part of #237


stephen-cox:
Just realised we might want to get #209 merged before doing a 2.0.0 release.

stephen-cox:
Think we also want to get #228 in before this is merged

finnlewis:
@stephen-cox I'm happy for this to be merged once #209 is merged. 

Then we can have a 2.0.0 release!

PR #250, opened 2021-05-26
Patch Drupal core to fix tests running with Paratest

### LINKED ISSUE #249: Tests somtimes fail when running with Paratest with "mkdir(): File exists"
This effects about 10% of test runs so it enough to be annoying. for an example see: https://github.com/localgovdrupal/localgov_step_by_step/runs/2674065876?check_suite_focus=true

Looks like there's a d.o.issue for this with patch: https://www.drupal.org/project/drupal/issues/3192365

stephen-cox:
See: https://www.drupal.org/project/drupal/issues/3192365

Closes #249

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,8 +48,9 @@
       },
       "patches": {
         "drupal/core": {
-          "Fix for Nested Paragraphs create multiple drag handles https://www.drupal.org/project/drupal/issues/3092181": "https://www.drupal.org/files/issues/2020-12-08/3092181-142-9X_0.patch",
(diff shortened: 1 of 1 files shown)


stephen-cox:
It's hard to test this as the problem only occurs with about 10% of test runs, but the issue on d.o. describes the issue we're seeing and at least one person reports the patch working.

finnlewis (approved):

stephen-cox:
Thanks @finnlewis - Merging

PR #272, opened 2021-06-18
Require LocalGov Alert Banner 1.1.0

stephen-cox:


andybroomfield (approved):
Looks good to me.
Will this make the profile 2.1 or 2.0.1?

stephen-cox:
I'll create a PR for a  2.0.1 release and we can discuss whether this should be 2.1.0

PR #271, opened 2021-06-16
Remove patch from https://www.drupal.org/project/drupal/issues/309218…

finnlewis:
…1#comment-13945784 now that this has been committed https://www.drupal.org/project/drupal/issues/3083051

PR #274, opened 2021-06-22
2.0.1 release

stephen-cox:
Includes LocalGov Alert Banner version 1.1.0.

Changelog:

#250 @stephen-cox Patch Drupal core to fix tests running with Paratest
#271 @finnlewis Remove patch for Nested Paragraphs create multiple drag handles 
#272 @stephen-cox Require LocalGov Alert Banner 1.1.0


finnlewis (approved):
Not tested locally, but tests pass and it these commits have been approved already. It's a 👍 from me!

andybroomfield (approved):
Ok to add this, are we agreed this is 2.0.1 and not 2.1?

stephen-cox:
We probably need some guidelines on when a minor release of the profile is required. There will be a release including sitewide search soon. I suggest this should be a minor release rather than a patch release, but let's discuss on that PR once it's created. Doing a 2.0.1 release for now.

andybroomfield:
@stephen-cox backport to 1.x also required?

PR #322, opened 2021-08-23
Set LocalGov Base as default theme

### LINKED ISSUE #317: LocalGov Base theme - set as default theme in profile.
On a default install, the default theme should be localgov_base theme.

- [x] Do non-beta release (https://github.com/localgovdrupal/localgov_base/issues/119)
- [x] Update block settings for other modules (potentially tedious - with a load of PRs)
- [x] Set localgov_base as the default theme in the install profile.

stephen-cox:
- Sets the sets the default theme to LocalGov Base
- Updates blocks so they are positioned correctly with new theme
- Moves tabs block to tabs region


stephen-cox:
Before this PR can be merged the last commit (aliasing branches in the `composer.json` file) needs to be undone.

finnlewis (approved):
Looks great!

PR #336, opened 2021-09-06
Add the patch for unpublished entity references

### LINKED ISSUE #334: Entity Reference: unpublished items
Editors can't reference unpublished nodes; when creating content it is a common requirement to create both overview and linked pages while unpublished.

Modules certainly affected:
 * LocalGov Services (landing page)
 * LocalGov Guides (overview)
 * LocalGov Step-By-Step (overview)

There is the additional factor that this ideally should also work for content in workflow, ie an editor can add content that's not yet approved. Lets restrict this to the current active revision (ie. what entity reference field references).

Fix suggested in guides localgovdrupal/localgov_guides#71 uses the patch from core issue https://www.drupal.org/project/drupal/issues/2845144
This works at least on the permissions to see unpublished content, if not workflow.

Suggested To Do:

 - [ ] Agree the core patch fixes at least part of the problem now, and assuming it never gets merged won't have any long term complications;
 - [ ] Test patch against all three modules, especially checking:
   - How referenced unpublished content is displayed to users that can see it;
   - How referenced unpublished content is (not) displayed to users that can't see it;
   - Which users can reference unpublished content;
   - What happens to users editing content they can't reference.

- [ ] write tests for other modules
- [ ] document odd behaviour where known

The patch won't fix all the issues, but they should be documented; and follow-up issue as appropriate (particularly for workflow?)

Related issues:
 *  localgovdrupal/localgov_services#71
 * localgovdrupal/localgov_step_by_step#24
 * localgovdrupal/localgov_guides#44
 * localgovdrupal/localgov_guides#27
 * localgovdrupal/localgov_guides#36
 * localgovdrupal/localgov_guides#40
 * localgovdrupal/localgov_guides#71

andybroomfield:
Fix #334

Draft of adding the patch to the profile

TODO: Add the test.
https://github.com/localgovdrupal/localgov_guides/pull/71/files

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,8 @@
             "drupal/core": {
                 "Fix for narrow forms on node/add screens, which is included in 9.2. See https://www.drupal.org/project/drupal/issues/3184667": "https://www.drupal.org/files/issues/2021-04-16/3184667.26.patch",
                 "Fix tests running with paratest. See https://www.drupal.org/project/drupal/issues/3192365": "https://www.drupal.org/files/issues/2021-01-12/3192365-3.patch",
-                "Fix Undefined index 'type' in Drupal\\Core\\Render\\Element\\Date::processDate() #3189302": "https://www.drupal.org/files/issues/2021-04-07/3189302_11.patch"
+                "Fix Undefined index 'type' in Drupal\\Core\\Render\\Element\\Date::processDate() #3189302": "https://www.drupal.org/files/issues/2021-04-07/3189302_11.patch",
+                "Users can't reference unpublished content even when they have access to it. See https://www.drupal.org/project/drupal/issues/2845144": "https://www.drupal.org/files/issues/2021-06-09/2845144_52.patch"
             },
             "drupal/gin_login": {
                 "Fix schema for gin_toolbar, see https://www.drupal.org/project/gin_login/issues/3192526": "https://www.drupal.org/files/issues/2021-01-13/gin_login-config_schema-3192526-8.patch"
--- a/tests/src/Kernal/UnpublishedParentOrChildTest.php
+++ b/tests/src/Kernal/UnpublishedParentOrChildTest.php
@@ -1,0 +1,104 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\localgov\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\Tests\node\Traits\NodeCreationTrait;
+use Drupal\Tests\user\Traits\UserCreationTrait;
+
+/**
+ * Entity reference tests.
+ *
+ * We should be able to reference unpublished Guide overview or Guide pages from
+ * each other.  This fails due to a core bug but taken care of by a patch.
+ *
+ * @see https://www.drupal.org/node/2845144
+ */
+class UnpublishedParentOrChildTest extends KernelTestBase {
+
+  use NodeCreationTrait;
+  use UserCreationTrait;
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  protected static $modules = [
+    'system',
+    'filter',
+    'field',
+    'text',
+    'options',
+    'user',
+    'node',
+    'localgov_roles',
+    'localgov_core',
+    'localgov_guides',
+  ];
+
+  /**
+   * {@inheritdoc}
+   *
+   * Sets up:
+   * - A user with guide related permissions of the localgov_editor role.
+   */
+  protected function setUp(): void {
+
+    parent::setUp();
+
+    $this->installSchema('system', ['sequences']);
+    $this->installSchema('node', ['node_access']);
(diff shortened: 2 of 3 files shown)


Adnan-cds:
Thanks Andy.  How about keeping the patch in this pull request and leaving the test in localgov_guides? @andybroomfield @ekes 

By the way, the indentation in composer.json may have been disturbed :(

Adnan-cds:
How about addressing the simplest use case (referring to unpublished pages) in this ticket and then treating content moderation in a separate ticket?  If we do that, then this change is already ready for use. @andybroomfield @ekes @tom-steel @richardaclarke

finnlewis:
@andybroomfield is hoping to get tests written ready for merge monday.

finnlewis:
@ekes could you please add issues for where this will have a UX impact on the content designer?

andybroomfield:
Noting this error
````
2x: Relying on entity queries to check access by default is deprecated in drupal:9.2.0 and an error will be thrown from drupal:10.0.0. Call \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with TRUE or FALSE to specify whether access should be checked. See https://www.drupal.org/node/3201242
    2x in UnpublishedParentOrChildTest::testUnpublishedParentAndChildSelection from Drupal\Tests\localgov\Kernel
````

andybroomfield:
I've added the Guide test and a new Step by Step test to this PR.
I tried adding Services Landing the same way, but this proved to be complicated as so many other modules and schemas need to be added to setup to work as a kernel test it becomes difficult (When writing own tests based on localgov_services, I usually bite the bullet and write functional browser tests).

I'm not sure there actully is much point writing tests for this against all modules. The only thing this is doing is adding the patch so non admin users can reference unpublished entities (Logged in as user 1, the issue does not present itself).
All we are doing in writing these tests is proving the patch applies. The patch itself includes tests and I worry we could end up spending too much time duplicating work and essentially testing core. We really only need to know that a non admin user can reference an unpublished node, and that could be done with an arbitury content type, though here as its the profile we have gone ahead and used Guides and Step by Step.

In other words, I don't belive not having a test for Services should block us reviewing and merging this.

andybroomfield:
The following is now a consistent fail (don't think its the test environment)

> 1) Drupal\Tests\localgov\Functional\LocalGovUpdateTest::testUpdate
Behat\Mink\Exception\ElementTextException: The text "Appointments with adult social care staff are now happening on the telephone." was not found in the text of the element matching css "main article .alert-info".

andybroomfield:
This is the code block that fails (in the update test).
Can someone check if this is to do with any other change in the profile (demo content / new theme)? 
Or would the patch affect this (I not sure how though).
/cc @stephen-cox @ekes @Adnan-cds 
````
    // Test Step by step overview page: Request support for an adult.
    $this->drupalGet('/adult-health-and-social-care/step-by-step/request-support-adult-step-step');
    $this->assertSession()->elementTextContains('css', 'header h1', 'Request support for an adult: step by step');
    $this->assertSession()->elementTextContains('css', 'main article .alert-info', 'Appointments with adult social care staff are now happening on the telephone.');
    $this->assertSession()->elementTextContains('css', 'main article .callout-danger', 'For urgent help');
    $this->assertSession()->elementTextContains('css', '.step-list', 'Find out what support we offer');
    $this->assertSession()->elementTextContains('css', '.step-list', 'Contact the support team');
    $this->assertSession()->elementTextContains('css', '.step-list', 'Apply for financial support');
    $this->assertSession()->elementTextContains('css', '.step-list', 'Choosing how to manage your care');
````

finnlewis:
Discussing this with @stephen-cox and @andybroomfield on a call now. 

@stephen-cox would like to co-ordinate the releases of this change with releases of other modules  like step-by-step, guides, services is needed. 

For this specific merge request, 
 - fix the tests. 
 - liaise with other modules to co-ordinate the releases

From my manual tests, a user 1 can already create unpublished guide pages and reference them from the guide overview. Next and previous work as expeted. 

A user 1 can already create unpublished step-by-step pages and reference them from the guide overview. Next and previous work as expeted. 

So it looks like entity reference is taking care of the access control. 

I'm happy fo this to be merged once tests pass and we can pick up and extend tests on step-by-tep and guide pages and services pages once this is in.

stephen-cox:
> This is the code block that fails (in the update test). Can someone check if this is to do with any other change in the profile (demo content / new theme)? Or would the patch affect this (I not sure how though).

Fix for failing test is here: https://github.com/localgovdrupal/localgov/pull/387

finnlewis (approved):
Tests pass and we discussed in our 'merge monday' meeting yesterday, so happy for this to merge and unblock https://github.com/localgovdrupal/localgov_guides/pull/71 (cc @Adnan-cds )

Adnan-cds:
> In other words, I don't believe not having a test for Services should block us reviewing and merging this.

Agree.

finnlewis:
Sounds like we're all good on this on @andybroomfield !

PR #337, opened 2021-09-07
Update Admin Toolbar to 3.x branch

MattOz-CDS:
**Observations**

- I noticed that when admin_toolbar_search is enabled and Gin is the administration theme, the search input only displays on the admin index page.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


MattOz-CDS:
This is more or less identical to what was approved on the 2.x branch so I've merged  this as well.

PR #366, opened 2021-09-22
Use carets instead of tildes for our version constraints to allow for…

### LINKED ISSUE #365: Version requirements for localgov profile's dependencies are too strict
Hello,
I was trying to update the localgov_alert_banner module in a fresh localgov test site but it wouldn't budge.  Turned out even though the latest version for localgov_alert_banner is 1.2.0, the localgov profile's version [requirement demands](https://github.com/localgovdrupal/localgov/blob/4d6720cf7d068d193e32550e31e6913a08a31a05/composer.json#L22) ```~1.1.0``` which I feel is too strict.  The same is true for the other dependencies of the localgov profile. ```~1.1.0``` is equivalent to ```1.1.x >= 1.1.0``` whereas we may need something like ```1.x.x >= 1.1.0``` which ```^1.1.0``` can satisfy easily.  So I am wondering if we should replace all the tildes with carets in the localgov profile's composer.json file.

@finnlewis @danchamp @andybroomfield @stephen-cox @ekes @MattOz-CDS

finnlewis:
… future updates with non-breaking changes.


ekes:
My tendency would be to leave it, so you have to make explicit the feature releases (and any dependency changes) in releases of localgov.

MattOz-CDS (approved):
Thanks Finn

Boosmith:
@finnlewis I would support it. Carats are much more widely used in the community. Tildes could make updating too fiddly

finnlewis:
Discussed in tech group dropin - @ekes @stephen-cox @Adnan-cds to approve before merging.

stephen-cox (approved):
Looks to be pulling the latest releases of all modules which is what we want :+1:

finnlewis:
Thanks @stephen-cox 
I'll leave this for another day in case @ekes  and @Adnan-cds can confirm too.

Adnan-cds:
> I'll leave this for another day in case @ekes and @Adnan-cds can confirm too.

Sorry, totally ran out of time reviewing [Eke's Directory sorting](https://github.com/localgovdrupal/localgov_directories/pull/152) change.  I don't want to hold this one back.  If Steve is happy, I am happy.

ekes:
Reservations remain, but happy to go with the flow.

Adnan-cds:
> Reservations remain, but happy to go with the flow.

Yeah, let's keep an open mind on this.  I noticed that while Drupal core uses carets, it also ships a [composer.lock file](https://git.drupalcode.org/project/drupal/-/blob/9.3.x/composer.lock) to keep tight control over what gets used.  If carets prove problematic later then we may have to move back to tildes.

PR #380, opened 2021-10-01
Set default theme to localgov_scarfolk and config for block placement…

### LINKED ISSUE #379: LocalGov Scarfolk sub-theme - set as default on install
The default install of the localgov profile currently sets the localgov_base theme as the default theme.

The localgov_base theme is meant to be a minimal starting point and does not include 'nicer' webfonts. 

Any council using the distribution is expected to create a sub theme and customise to their own end. 

We want to show the distribution off in the best light for testing, so I think it makes sens to set the default theme to localgov_scarfolk.

finnlewis:
… on both localgov_scarfolk and localgov_base.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "localgovdrupal/localgov_services": "^2.0.0",
         "localgovdrupal/localgov_step_by_step": "^2.0.0",
         "localgovdrupal/localgov_subsites": "^2.0.0",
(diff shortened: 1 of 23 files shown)


stephen-cox (approved):
I have checked out all the related branches and run an install of the LocalGov profile and the demo content module and largely looks to be working fine; just a couple of issues with the Directories module: https://github.com/localgovdrupal/localgov_directories/pull/159#pullrequestreview-771381450

PR #384, opened 2021-10-04
Bump version of localgov_alert_banner to ~1.2.0.

finnlewis:


andybroomfield (approved):
Tests are passing for 1.x, should now unblock some further PRs that need to test against 1.x.

Issue #396, opened 2021-11-04
Search api updates are failing (due to a patch?)


andybroomfield:
After doing composer update and trying to run updates with `lando drush updb -y` I get
> Fatal error: Cannot redeclare search_api_update_8107() (previously declared in /app/web/modules/contrib/search_api/search_api.install:341) in /app/web/modules/contrib/search_api/search_api.install on line 409

There is a duplicated search_api_update_8107() function, but it is not included in the vanillia 1.21 package which I downloaded from drupal.org.
The function does not appear anywhere in the vanilla search API module so I assume this is coming from a patch which needs to be updated.

````php
/**
 * Update Search API views to use the correct query type.
 *
 * In some cases, Views creates Search API views with the default "views_query"
 * query type instead of "search_api_query".
 */
function search_api_update_8107() {
  $config_factory = \Drupal::configFactory();
  $changed = [];

  foreach ($config_factory->listAll('views.view.') as $view_config_name) {
    $view = $config_factory->getEditable($view_config_name);
    if (substr($view->get('base_table'), 0, 17) == 'search_api_index_') {
      $displays = $view->get('display');

      $updated = FALSE;
      foreach ($displays as $id => $display) {
        if (($display['display_options']['query']['type'] ?? '') === 'views_query') {
          $displays[$id]['display_options']['query']['type'] = 'search_api_query';
          $updated = TRUE;
        }
      }

      if ($updated) {
        $view->set('display', $displays);
        // Mark the resulting configuration as trusted data. This avoids issues
        // with future schema changes.
        $view->save(TRUE);
        $changed[] = $view->get('id');
      }
    }
  }

  if (!empty($changed)) {
    $vars = ['@ids' => implode(', ', array_unique($changed))];
    return t('The following views have been updated to use the "search_api_query" query type instead of "views_query": @ids.', $vars);
  }

  return NULL;
}
````

ekes:
Oh! We've got a patch that includes an update_X hook :-( `<kicks self>`

This is going to be 'fun'.

I guess we need to drop the update index by one, for anyone we're sure has run the patch hook? Then maybe it can be changed into a named post-update, that would be safer. That or in a custom module with something to take account of the code getting merged. Otherwise this will happen again I'd guess.

andybroomfield:
Thanks for looking @ekes do you know which module is including the patch and is it a patch from drupal.org?

andybroomfield:
[...]

PR #397, opened 2021-11-08
Pin search api to 1.20

andybroomfield:
To test if this fixes failing search api patch.


finnlewis (approved):
Let's do it! This is a good temporary fix to allow us to continue working while we work out https://github.com/localgovdrupal/localgov/issues/396

PR #403, opened 2021-11-10
Patch Drupal core to fix breadcrumb assertions

### LINKED ISSUE #402: Breadcrumb assertions fail when not using Classy
AssertBreadcrumbTrait::getBreadcrumbParts expects the breadcrumb wrapper to have a specific CSS class. This was the case when using the classy (by default) however tests now need to specify the `$defaultTheme` and best practise is to avoid depending on markup provided by core. 

See https://www.drupal.org/node/3083055

j4-m:
AssertBreadcrumbTrait::getBreadcrumbParts expects the breadcrumb wrapper to have a specific CSS class, when this isn't the case the test fails even if the breadcrumb exists. This patch tries to match the breadcrumb by looking for an aria attribute which should means tests can pass on most themes.

See https://www.drupal.org/project/drupal/issues/3248309.

Closes #402


finnlewis (approved):
Thanks @j4-m - looks good to me, let's merge this

PR #405, opened 2021-11-13
Work round skipped search_api update, add new patch for search_api schema fix: #396

ekes:
https://github.com/localgovdrupal/localgov/issues/396

 * Rather than try and play with schema numbering as it was this profile
   that included the search_api patch, this module can run the
   search_api update. It runs before search_api further updates.
 * The new search_api patch, includes tests, and uses a post_update
   hook. If Thomas still moves it back to and update hook that's fine as
   it can be run more than once.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "drupal/gin_login": "^1.0@RC",
         "drupal/gin_toolbar": "^1.0@beta",
         "drupal/simple_sitemap": "^3.7",
-        "drupal/search_api": "1.20",
+        "drupal/search_api": "^1.21",
         "drush/drush": ">=10",
         "localgovdrupal/localgov_alert_banner": "^1.2.0",
         "localgovdrupal/localgov_base": "^1.0.0",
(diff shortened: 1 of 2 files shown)


finnlewis:
Just discussing at Merge Monday, and this looks good to me. 

@andybroomfield do you want to take a look before we merge ?

andybroomfield:
@finnlewis On leave this week, but if it fixes the issue and tests then I think ok. I take it I have to manually rollback the search api schema to apply this?

andybroomfield:
We have dedicated branch now, but I think it would be useful to have the Drupal 9 support mainlined.

ekes:
> I take it I have to manually rollback the search api schema to apply this?

If you've already upgraded to search_api it will just have skipped 8107, but run 8108 onward. This patch does the work of search_api_update_8107 in localgov_update_8201. So assuming it's fine that they ended up in a different order it will work for you even if you've already updated. For people who didn't update localgov_update_8201 runs before search_api_update_8108. And the work that was done in the patched search_api_update_8107 has moved into a post_update hook (and it's fine if that - and actually the real 8107 - are run multiple times).

finnlewis (approved):
Great, happy with this, logical and tests pass. Let's do it!

ekes:
Yes if the upgrade test passes, and there's no schema error, that's pretty much what this is fixing.

PR #420, opened 2022-01-24
Fix deprecated code errors for Search API TaskStorageSchema class 

### LINKED ISSUE #419: Deprecated code failures
```
 ------ -------------------------------------------------------------- 
  Line   profiles/contrib/localgov/localgov.install                    
 ------ -------------------------------------------------------------- 
  34     Fetching class constant class of deprecated class             
         Drupal\search_api\Entity\TaskStorageSchema:                   
         in search_api:8.x-1.23 and is removed from search_api:2.0.0.  
           There is no replacement.                                    
  35     Fetching class constant class of deprecated class             
         Drupal\search_api\Entity\TaskStorageSchema:                   
         in search_api:8.x-1.23 and is removed from search_api:2.0.0.  
           There is no replacement.                                    
 ------ -------------------------------------------------------------- 
```

stephen-cox:
Closes #419

### DIFF
--- a/localgov.install
+++ b/localgov.install
@@ -6,7 +6,6 @@
  */
 
 use Drupal\user\Entity\User;
-use Drupal\search_api\Entity\TaskStorageSchema;
 
 /**
  * Implements hook_install().
--- a/localgov.install
+++ b/localgov.install
@@ -26,13 +26,9 @@ function localgov_install() {
  * Run the potentially skipped search_api_update_8107() safely.
  */
 function localgov_update_8201() {
-  $module_handler = \Drupal::moduleHandler();
-  if ($module_handler->moduleExists('search_api')) {
-    $manager = \Drupal::entityDefinitionUpdateManager();
-    $entity_type = $manager->getEntityType('search_api_task');
-    if ($entity_type->getHandlerClass('storage_schema') != 'TaskStorageSchema') {
-      $entity_type->setHandlerClass('storage_schema', 'TaskStorageSchema');
-      $manager->updateEntityType($entity_type);
-    }
-  }
+  // Superseded by search_api_update_8110().
+  //
+  // Sites that have run this update, will get updated by 8110.
+  // Sites that have not run this update will also get 8110 and shouldn't risk
+  // running this version after search_api_update_8110.
 }
--- a/localgov.install
+++ b/localgov.install
@@ -31,8 +31,8 @@ function localgov_update_8201() {
   if ($module_handler->moduleExists('search_api')) {
     $manager = \Drupal::entityDefinitionUpdateManager();
     $entity_type = $manager->getEntityType('search_api_task');
-    if ($entity_type->getHandlerClass('storage_schema') != TaskStorageSchema::class) {
-      $entity_type->setHandlerClass('storage_schema', TaskStorageSchema::class);
+    if ($entity_type->getHandlerClass('storage_schema') != 'TaskStorageSchema') {
+      $entity_type->setHandlerClass('storage_schema', 'TaskStorageSchema');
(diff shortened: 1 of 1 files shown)


finnlewis:
@ekes - would you be able review this and confirm if this will still work around as intended? Thanks!

localgov.install
```diff
@@ -31,8 +30,8 @@ function localgov_update_8201() {
   if ($module_handler->moduleExists('search_api')) {
     $manager = \Drupal::entityDefinitionUpdateManager();
     $entity_type = $manager->getEntityType('search_api_task');
-    if ($entity_type->getHandlerClass('storage_schema') != TaskStorageSchema::class) {
-      $entity_type->setHandlerClass('storage_schema', TaskStorageSchema::class);
+    if ($entity_type->getHandlerClass('storage_schema') != 'TaskStorageSchema') {
```
Adnan-cds: Hi Steve, the search_api module's [search_api_update_8110()](https://git.drupalcode.org/project/search_api/-/blob/e8bb207776db70b8a35768ad3b2a59bbe31d7149/search_api.install#L406) has taken a slightly different approach.  I don't know if that's relevant here but thought I will mention it anyway. @stephen-cox

ekes: Hi, sorry had seen this on.

Adnan well spotted that.

The full patch https://git.drupalcode.org/project/search_api/-/commit/e8bb207776db70b8a35768ad3b2a59bbe31d7149 deprecates the TaskStorageSchema class completely.

stephen-cox: Happy for someone to suggest the changes required here; I'm flat out and not likely to have time to look at this for a few days and tests are failing until this gets fixed.

ekes: Having deprecated code in a install file seems fine to me. Like it will need to be removed and a major version update made when the code goes. But this is the same update that search api itself also still has https://git.drupalcode.org/project/search_api/-/blob/e8bb207776db70b8a35768ad3b2a59bbe31d7149/search_api.install#L341 we needed to run it because of that blunder move of patching an update.

Search API will then in its updates do it's next update https://git.drupalcode.org/project/search_api/-/blob/e8bb207776db70b8a35768ad3b2a59bbe31d7149/search_api.install#L406

Or is the problem that localgov updates are run _after_ search api's on installs that haven't yet run localgov_update_8201(). So localgov_update_8201() undoes search_api_update_8110()? If that's the case... then we should rewrite history of the updates; swapping it for the same as search_api_update_8110() would seem 'harmless' (in the I don't think it digs us further into the hole).

ekes: So just on the hole digging thing. To run it past everyone:-

Some people will have run localgov_update_8201() and should have run some search_api updates < 8110 at which point they will run 8110 with the next update - no problem for them.
Other people won't have run localgov_update_8201 at which point they'll also be either search_api updates < 8107, or search_api_updates with whatever we patched as 8107. Running a localgov_update_8201 and also running all the search api updates from either before 8107 or from 8108 is _safe_?
Can we get a couple of agreements on that last sentence. Then we can change it?

ekes:
So I ran through the logic I mentioned above, and realized we don't actually want to run the content of the update.

We've not done anything to cause search_api_update_8110() to be skipped, unlike 8107. 8110 changes the work both of this update localgov_update_8201 and its equivalent in search api search_api_update_8107. So we don't need to run it on sites that haven't yet got the update, more so, we really don't want to run it after then new search_api_update_8110. Sites that have got our version will also get the new 8110 and everyone is happy.

Adnan-cds:
Code change looks good to me.  Thanks :)

Adnan-cds (approved):
Looks good.  Thanks :)

PR #436, opened 2022-04-11
Test adding cron schedule and adding the PROJECT_NAME for schedule ru…

### LINKED ISSUE #435: Ensure full integration tests run on the install profile.
Noting that we are moving to running tests on each repo against the code within that project.

We then want to schedule tests to run daily on the profile repo.

finnlewis:
…ns which do not have an event to access the repo name.


stephen-cox (commented):
The LocalGov project runs all tests as it contains no tests itself, so I think it makes sense to add the nightly cron trigger there. I've created an issue for it there: https://github.com/localgovdrupal/localgov_project/issues/79

PR #459, opened 2022-07-07
Release 2.3.11

finnlewis:


finnlewis:
The urgency with this release is the removal of the patch for Search API https://github.com/localgovdrupal/localgov/pull/459/commits/9307e7a1198672a66ceebdc2e039c53d819bd2c7

The tests are failing for other reasons ( schema issues due to leaflet module update is it? )https://github.com/localgovdrupal/localgov/runs/7235635056?check_suite_focus=true#step:5:23

So @ekes and @stephen-cox were discussing getting this out and fixing that soon.

Any critical concerns with that @Adnan-cds @andybroomfield @markconroy @stephen-cox @ekes ?

andybroomfield:
If this is just removing a patch that is included in the update, then let's push it.
However can we add to the release notes that the patch has been removed and maybe up the minimum of search_api to the one that includes the patch.

Should we also consider not having exit on patch failure or is there a reason to keep it?

markconroy:
I'm happy with us getting this released if it's just removing a patch that is no longer needed.

andybroomfield:
I assume the update to search_api should also clear this issue?
> PHP Fatal error:  Cannot redeclare search_api_post_update_views_query_type() (previously declared in /Users/Andy/Sites/bhcclocalgov/docroot/modules/contrib/search_api/search_api.post_update.php:28) in /Users/Andy/Sites/bhcclocalgov/docroot/modules/contrib/search_api/search_api.post_update.php on line 74

stephen-cox (approved):
Looks like the test fixes are going to take a while, so let's get this released!

PR #462, opened 2022-07-13
adding YML templates to localgov repo

LauraHilliger:
created two YML templates

PR #464, opened 2022-07-18
adding YML templates to localgov repo

msayoung:


msayoung:
@LauraHilliger tried to detangle the commits we wanted and those we didn't from the PR here: https://github.com/localgovdrupal/localgov/pull/462 but found it easiest to recreate your commit and have credited you. Can you check it looks right? 

cc @aaronhirtenstein

LauraHilliger:
@msayoung - THANK YOU! Looks good to me! Hopefully they work 😬 . If these are merged we need to remember that when @aaronhirtenstein and the tech crew decide on some labels, you/they/we might need to edit the auto-labeling on these templates.

PR #475, opened 2022-10-19
Added instructions for using DDEV as a local dev environment.

siliconmeadow:


finnlewis:
Thanks for this @siliconmeadow 

There is already some documentation hidden here: https://docs.localgovdrupal.org/devs/getting-started/working-with-ddev.html

I think it would be good to avoid duplication by cross linking from the README to the docs and vice versa. 

Also noting that the DDEV documentation can only be found through search at present, so we want to add a link to it from the documentation menu on the left hand side.

finnlewis:
@msayoung any thoughts on how to tie this all together?

msayoung:
Firstly thanks for this @siliconmeadow 

I suggest we move this over to the docs site, extending the documentation Finn pointed to.  https://docs.localgovdrupal.org/devs/getting-started/working-with-ddev.html
We can link to the DDEV page from https://docs.localgovdrupal.org/devs/getting-started/ 

We can then link to that documentation from the README here.

Does that make sense?

finnlewis:
@siliconmeadow might you have time to look at this?

finnlewis:
@siliconmeadow we now have: https://docs.localgovdrupal.org/devs/getting-started/working-with-ddev.html

Shall we close this ?

finnlewis:
Also see https://github.com/localgovdrupal/localgov_project/pull/102

finnlewis:
@siliconmeadow confirmed on Slack that we can close this one

PR #485, opened 2023-01-11
Feature/466 autosave

### LINKED ISSUE #466: Autosave content
In the process of moving to LocalGov Drupal, one thing we're missing from our current CMS is autosave functionality.

[localgov_microsites](https://github.com/localgovdrupal/localgov_microsites/) is using `drupal/autosave_form` - could we consider including this in core LocalGov Drupal? Would also welcome any other alternatives.

finnlewis:
@keelanfh @markconroy @andybroomfield 

Here's a PR to test, just adds and installs the module with default configuration (saves content forms every 60 seconds)

Coming from https://github.com/localgovdrupal/localgov/issues/466

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     ],
     "require": {
         "drupal/admin_toolbar": "^3.0",
+        "drupal/autosave_form": "^1.3",
         "drupal/core": "^9.1",
         "drupal/devel": "^4.1",
         "drupal/gin": "^3.0@alpha",
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "drupal/admin_toolbar": "^3.0",
-        "drupal/autosave_form": "^1.3",
+        "drupal/autosave_form": "^1.4",
         "drupal/core": "^9.1",
         "drupal/devel": "^4.1",
         "drupal/gin": "^3.0@alpha",
--- a/localgov.info.yml
+++ b/localgov.info.yml
@@ -35,6 +35,7 @@ install:
   # Contrib
   - admin_toolbar:admin_toolbar
   - admin_toolbar:admin_toolbar_tools
+  - autosave_form:autosave_form
   - entity_browser:entity_browser
   - simple_sitemap:simple_sitemap
   # LocalGov Drupal


andybroomfield:
Initial look through.
Installing this branch does not appear to enable autosave by default in an existing install.
Once enabled, it can still be removed.
I noticed it doesn't come with any configuration so to be useful is going to need to be enabled and set to autosave on all content types, which I guess also means there is going to be something needed to update that, which brings back the issues in the original PR about how to make this opt in.

keelanfh:
@andybroomfield If you don't select any of the content types, it defaults to being enabled for all content types. So I don't think that would require ongoing changes.

markconroy:
Would be good to have a notification or something to existing sites via `drupal_set_message` to say "Hey, we've added an autosave module, would you like to enable it".

We've lots of nice new features that existing sites don't get by default. Sorry, I know this is a side issue to the main PR.

finnlewis:
No need to apologise Mark, this was a large topic of conversation in our [Tech Group Governance meeting yesterday](https://docs.google.com/document/d/1HxtUig60ly91yIhlNPIooN4nnSNWIdIE0lUT_4qIt_A/edit#heading=h.mxireog8qobo). 

Specifically: "What is the policy for adding modules to default install LocalGov eg. autosave"

I will draft a policy for review, but key questions include: 

1. are there at least 2 councils requesting a new module or feature? 
2. Are users able to optionally disable the module / feature?
3. Are most of our users using / wanting the module / feature?
4. Should we check with the product group / product owner?
5. How do we communicate new potential additions to existing instalations? (Release notes / README / drupal_set_message like you suggest @markconroy ?)

So perhaps we can use this pull request as a case in point to draft and agree a working policy for such things.

andybroomfield:
Thanks @keelanfh I did notice that after a while, I got thrown as autosave does not work for /node/add, so only noticed it when I went back to edit.
I'd worry though about rolling it out to all content, to be useful it may welll need to have some default config shipped with it, which brings us back to how to make this opt in.

andybroomfield:
Worth putting this to draft whilst the discussion is ongoing @finnlewis ?

finnlewis:
I've started a draft policy for discussion on the general questions this raises: https://github.com/localgovdrupal/localgov/wiki/%5BDraft%5D-Policy-for-adding-a-new-feature,-functionality-or-modules-to-LocalGov-Drupal

keelanfh:
@andybroomfield that's interesting that it doesn't work for /node/add, I was not aware of that.

andybroomfield:
Going to need at least these patches on core I think to get it to work with paragraphs
https://www.drupal.org/project/drupal/issues/2824097
https://www.drupal.org/project/drupal/issues/2844229

And this patch on the module to get autosave on node creation
https://www.drupal.org/project/autosave_form/issues/2924006

finnlewis:
This is still being discussed in the backlog group with some updates being tested.

finnlewis:
Still being discussed https://github.com/localgovdrupal/localgov/issues/466

finnlewis:
@andybroomfield mentioned 
1. this not working with paragraphs... but this is installed in Drupal CMS so maybe this is resolved.
2. need to be able to disable it for performance (Acquia)

PR #486, opened 2023-01-12
Fix two small typos

alexwybraniec:


finnlewis (approved):
Thanks @alexwybraniec !

PR #492, opened 2023-01-25
Enable stable9 theme if localgov_base theme is enabled 

stephen-cox:
Part of localgovdrupal/localgov_base#347

### DIFF
--- a/localgov.install
+++ b/localgov.install
@@ -32,3 +32,13 @@ function localgov_update_8201() {
   // Sites that have not run this update will also get 8110 and shouldn't risk
   // running this version after search_api_update_8110.
 }
+
+/**
+ * Enable stable9 theme if localgov_base theme is enabled.
+ */
+function localgov_update_9501() {
+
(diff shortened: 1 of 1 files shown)


markconroy (commented):
I'd probably remove the `if` statement here and just enable stable9 across the board. I don't see any downsides to it.

stephen-cox:
I'm not sure enabling stuff for everyone that might not be needed is a good thing and I think the condition catches everyone using a `localgov_base` based theme with the distribution. They will need localgov_base enabled and so will get stable9 enabled . Happy to discuss though.

finnlewis (approved):
This works for me locally, thanks @stephen-cox !
https://github.com/localgovdrupal/localgov_base/pull/347 then also works nicely without errors. 
Phew!

stephen-cox:
Discussed at Merge Monday and the preference is to only enable stable9 when needed and so to keep the condition.

PR #517, opened 2023-03-24
Feature/166 date formats

### LINKED ISSUE #166: Default date format
The default date formats (e.g. Default long date, Default medium date, Default short date) are set to Drupal's default which is based on the mm/dd/yyyy format.  Should we consider updating these to dd/mm/yyyy or leave these as they are in case LocalGov spreads beyond borders?

FAO @finnlewis @willguv

keelanfh:
Fixes #166 

This should introduce the formats discussed in that PR. This prevents American-style date formats from appearing in the content editor list views.

### DIFF
--- a/config/install/core.date_format.long.yml
+++ b/config/install/core.date_format.long.yml
@@ -1,0 +1,7 @@
+langcode: en
+status: true
+dependencies: {  }
+id: long
+label: 'Default long date'
+locked: false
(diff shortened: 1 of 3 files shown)


ekes (approved):
Provides sensible date formats for an initial install for the standard date formats.

I'm going to assume it's not worth pushing this update as people will have configured these formats to what they specifically want.

PR #530, opened 2023-04-19
Enable Gin Login and Toolbar modules as part of a site install 

### LINKED ISSUE #525: Enable Gin login and toolbar for new users
Following on from #518, the Backlog Group has agreed that:

- Gin login and toolbar should be enabled by default 
- We should communicate this feature to existing users, so they can enable it if they wish

stephen-cox:
Fixes #525


finnlewis (approved):
Yeah baby, works for me and feels lovely! :+1:

PR #550, opened 2023-05-25
Disable visitor account registration

### LINKED ISSUE #468: Anonymous users can register accounts
In the profile, anonymous users can register accounts 

image

Do we want to change this default

stephen-cox:
Fixes #468


finnlewis (approved):
I think this is wise, thanks @stephen-cox

andybroomfield (approved):

Issue #512, opened 2023-03-23
Change default destination screen after login


willguv:
On login, users are presented with a mostly blank screen with their name on

![Image](https://user-images.githubusercontent.com/708140/227202491-d1eec514-f221-42bc-b46a-5e77aa282f23.png)

Most users are content designers, and the screen they need is /admin/content

![Image](https://user-images.githubusercontent.com/708140/227202530-094b90ff-2746-438e-a64f-37e324209cdf.png)

Let's discuss if making this change disadvantages other types of users

(Separately, does this remove the issue of having two different look and feels for admin, shown in the screenshots above)

keelanfh:
I think the issue of having two different look and feels is still there - if a content designer clicks on any of those links (e.g. "Get Online at Home", they'll be back to the preview of the page (whether that's expected behaviour or not is another matter!) - and there are other ways of ending up back there too).

Nevertheless, I am very much in favour of this. Discoverability of the interface when a new user logs in for the first time is really important (think of how much work GOV.UK Notify did on this), and this really gets in the way.

willguv:
@keelanfh just added #518

willguv:
This has come up a few times now - mentioned to me and identified in Essex's admin usability research

@stephen-cox is there an easy fix here? Thanks

stephen-cox:
The current behavior is the Drupal default, but it should be easy enough to override this and redirect to the content overview page.

We would want to double check the user has permissions to manage content as it's possible to add users without this permission. In this case we could either keep the default redirect or redirect to the homepage instead.

ekes:
Just to point out the usual solution for this is designing the /user/{user} page to include links and lists (of content) that are likely to be what that particular user wants (based on their posts, or their role or even configurable).

stephen-cox:
> Just to point out the usual solution for this is designing the /user/{user} page to include links and lists (of content) that are likely to be what that particular user wants (based on their posts, or their role or even configurable).

That's a good point, and some sites might have already done this, so if we were to add such a redirect we would want it to be configurable so it wasn't enabled for existing sites by default.

willguv:
Thanks @ekes @stephen

My hunch is redirect for now (double checking for permissions), and write another card for later about redesigning the /user page

Let's hold fire and I'll check with Backlog Group

willguv:
@stephen-cox so we're saying:

- for users with the ability to edit or create content (which roles?), the default destination screen should be /admin/content

[...]

PR #556, opened 2023-06-08
Redirect users to content overview page at login

stephen-cox:
Part of #512

Description of the functionality here: https://github.com/localgovdrupal/localgov/issues/512#issuecomment-1582694200

### DIFF
--- a/modules/localgov_login_redirect/localgov_login_redirect.module
+++ b/modules/localgov_login_redirect/localgov_login_redirect.module
@@ -6,7 +6,6 @@
  */
 
 use Drupal\Core\Session\AccountInterface;
-use Drupal\Core\Url;
 
 /**
  * Implements hook_user_login().
@@ -32,9 +31,8 @@ function localgov_login_redirect_user_login(AccountInterface $account) {
   }
 
   // Set destination parameter to redirect user with access permissions.
-  $route = $config->get('redirect_route') ?? 'view.content.page_1';
-  $url = Url::fromRoute($route);
-  if ($url->access($account)) {
+  $path = $config->get('redirect_path') ?? '/admin/content';
+  if ($url = \Drupal::service('path.validator')->getUrlIfValid($path)) {
     $request->query->set('destination', $url->toString());
   }
 }
--- a/modules/localgov_login_redirect/localgov_login_redirect.module
+++ b/modules/localgov_login_redirect/localgov_login_redirect.module
@@ -13,14 +13,28 @@ use Drupal\Core\Url;
  */
 function localgov_login_redirect_user_login(AccountInterface $account) {
 
-  // Redirect user at login.
+  // Check login redirect is enabled.
   $config = \Drupal::configFactory()->get('localgov_login_redirect.settings');
+  if ($config->get('enabled') !== TRUE) {
+    return;
+  }
+
+  // Check user is using a login link.
+  $current_route = Drupal::routeMatch()->getRouteName();
+  if (in_array($current_route, ['user.reset', 'user.reset.login'])) {
+    return;
+  }
+
+  // Check for a destination parameter.
   $request = \Drupal::request();
-  if ($config->get('enabled') === TRUE && is_null($request->query->get('destination'))) {
-    $route = $config->get('redirect_route') ?? 'view.content.page_1';
-    $url = Url::fromRoute($route);
-    if ($url->access($account)) {
-      $request->query->set('destination', $url->toString());
-    }
+  if (!is_null($request->query->get('destination', NULL))) {
+    return;
+  }
+
+  // Set destination parameter to redirect user with access permissions.
+  $route = $config->get('redirect_route') ?? 'view.content.page_1';
+  $url = Url::fromRoute($route);
+  if ($url->access($account)) {
+    $request->query->set('destination', $url->toString());
   }
 }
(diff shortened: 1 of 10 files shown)


andybroomfield:
Out of interest, why in the profile and not localgov_core. Sounds like it may have utility outside the profile and I thought we wern't putting modules into profile?

modules/localgov_login_redirect/localgov_login_redirect.module
```diff
@@ -0,0 +1,26 @@
...
+ * Hooks for LocalGov Login Page module.
+ */
+
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Url;
+
+/**
+ * Implements hook_user_login().
+ */
+function localgov_login_redirect_user_login(AccountInterface $account) {
```
ekes, line 13: Is this going to be too blunt? If I look at https://git.drupalcode.org/project/login_destination/-/blob/8.x-2.x/login_destination.module#L46 they do stuff like check that it's not a one time login link to reset the password. This is in part because that module lets you configure these things, but also tells me it needs to be accounted for.

modules/localgov_login_redirect/src/Form/LoginRedirectSettingsForm.php
```diff
@@ -0,0 +1,115 @@
...
+    $form = parent::buildForm($form, $form_state);
+    $config = $this->config('localgov_login_redirect.settings');
+
+    $form['enabled'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Enable login redirect'),
+      '#description' => $this->t('Redirect users to the content overview page at login.'),
+      '#default_value' => $config->get('enabled'),
+    ];
+    $form['redirect_route'] = [
```
ekes: If this is to solve the primary request (redirect to content list) it makes sense; but to generalize the chance is it'll need arguments (like user id) so tokens or so?

ekes (commented):
It's beta, it's seeking maintainers, but reading the code of https://git.drupalcode.org/project/login_destination for this review it seems reasonably written and feature complete.

stephen-cox:
> Out of interest, why in the profile and not localgov_core. Sounds like it may have utility outside the profile and I thought we wern't putting modules into profile?

Was not aware of anything that said we weren't putting modules in the profile.

It can go anywhere but wasn't sure there would be wider use for this and putting it in the profile makes it slightly easier to enable at install time (the config and code lives in a single repo)

stephen-cox:
> It's beta, it's seeking maintainers, but reading the code of https://git.drupalcode.org/project/login_destination for this review it seems reasonably written and feature complete.

There are a few modules that do this sort of thing, but as this was a relatively simple requirement, I figured it was easier to create a module to do the redirect that has a simple config and is easy to turn on and off.

If there's a preference to use contrib here I can look at using a module for this. Login Destination seems to use config entities so should be easy enough to enable the module and install the necessary config.

modules/localgov_login_redirect/localgov_login_redirect.module
stephen-cox: Good catch - we don't want to redirect people when they're resetting passwords.

modules/localgov_login_redirect/src/Form/LoginRedirectSettingsForm.php
stephen-cox: Yep - good thought. Was keeping it simple. Not sure how much work it is to add token substitution to this, but can look into it if we're happy with the approach here.

stephen-cox: Yep - good thought. Was keeping it simple. Not sure how much work it is to add token substitution to this, but can look into it if we're happy with the approach here.

andybroomfield:
BHCC Opted for [Login And Logout Redirect Per Role](https://www.drupal.org/project/login_redirect_per_role) but that is because we need to redirect different groups of users to different places.
I see no problem (and its possibly a benefit) on having a simple module to handle the general redirect, and then reccomend other modules or ECA for more complex needs.

modules/localgov_login_redirect/localgov_login_redirect.module
stephen-cox: I've added a check and a test for password reset links.

modules/localgov_login_redirect/src/Form/LoginRedirectSettingsForm.php
stephen-cox: I'm not sure quite what would be required to allow tokens in the route. Is this required for a first iteration of this or can it be put off for now?

finnlewis:
@stephen-cox do you think we could change the input from a Drupal route to a path. 

image

Not sure if this helps: https://gist.github.com/GoZOo/6e484f4e3001242a91922a63cc4195cd

stephen-cox:
> @stephen-cox do you think we could change the input from a Drupal route to a path.
> 
> Not sure if this helps: https://gist.github.com/GoZOo/6e484f4e3001242a91922a63cc4195cd

Thanks @finnlewis that did the trick. It's now using the path rather than the route

finnlewis (approved):
Tested locally and works nicely with path now. Easy to disable on config or disable the module, so all good I reckon!
Thanks @stephen-cox !

PR #573, opened 2023-06-29
Update README.md

finnlewis:


keelanfh:
Why is lando removed here? The rest of the commands include lando, so looks a bit inconsistent (I don't use lando, so I don't know - just checking).

finnlewis:
The initial `composer create project` command is run without lando, as we don't have a lando file at that point. Once we have run the `composer create project` we have a .lando.yml file, so can start lando and run commands in lando for a more consistent environment. 

If you run that command with lando, it creates another project inside the original project.

keelanfh (approved):

keelanfh:
OK!

PR #580, opened 2023-06-30
[2.x] Add but don't enable the Content Lock module

### LINKED ISSUE #543: Block concurrent editing
This has been raised by @keelanfh and others. I'll confirm with the content group this need is common

Two users should not be able to edit content at the same time

Ideally...

- /admin/content and other list pages (eg /admin/content/alert-banner) should show content that is locked and name of user that's editing)

- there should be some way to override a lock

This module looks like a good fit: https://www.drupal.org/project/content_lock

stephen-cox:
Fixes #543

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     ],
     "require": {
         "drupal/admin_toolbar": "^3.0",
+        "drupal/content_lock": "^2.3",
         "drupal/core": "^9.1",
         "drupal/devel": "^4.1",
         "drupal/gin": "^3.0@alpha",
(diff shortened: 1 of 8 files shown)


finnlewis:
Stephen says this is not quite ready yet. 

Finn (me) asked about documentation. Should we add some notes to clarify there are optional modules included that can be enabled? README.md? Dev docs? Both?

ekes (approved):
Approve but merge #611 *first*. I don't think we should be doing development on 2.x first any more, but porting back from dev on 3.x. Certainly don't want to accidentally be adding features to 2.x without them already being added to 3.x

For the rest I've done a manual test, and automated tests, and merged with current 2.x.

I wondered about putting it in the profile (rather than core), but as it's mainly bring in the dependency in composer and the menu it seems a pretty profile-y thing.

PR #588, opened 2023-07-05
Release 3.0.0-alpha1

stephen-cox:


stephen-cox:
The update test is failing. I think this is because it's trying to update from 1.0.0 to 3.0.0-alpha1, but the database created with 1.0.0 won't install in 3.x. This being the case, we'll need to update the test DB to 2.x

stephen-cox:
The failing test should be fixed with https://github.com/localgovdrupal/localgov_directories/pull/296

PR #598, opened 2023-07-31
Fix update test

### LINKED ISSUE #596: Test failure in 3.x branch: The "localgov_geo" entity type does not exist.
The test for the D9 to D10 update fails with the message 'The "localgov_geo" entity type does not exist.' This is likely because the DB we're testing doesn't have the localgov_geo_update module enabled.

stephen-cox:
Fixes #596


finnlewis (approved):
Looks good to me!

PR #601, opened 2023-08-07
Feature/176 localgov generator tag

alexburrows:
Cursor_and_view-source_https___localgov-dev_ddev_site

I've added the `\Drupal::Version` which returns the specific version so we don't need to mess about for 9 or 10 or even 11 again. 

This is related to https://github.com/localgovdrupal/localgov_core/issues/176
So https://github.com/localgovdrupal/localgov_core/pull/177 can be closed as discussed its better to have in the profile.

### DIFF
--- a/localgov.profile
+++ b/localgov.profile
@@ -15,7 +15,7 @@ function localgov_page_attachments(array &$attachments): void {
 
     if ($name == 'system_meta_generator') {
       $tag = &$html_head[0];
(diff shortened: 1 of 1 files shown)


stephen-cox (approved):
Looks good. Keeping the Drupal version in the header should mean sites still identify as Drupal sites as well as LGD sites.

stephen-cox:
This will also need porting to the 3.x branch so it appears on D10 sites as well

PR #612, opened 2023-08-24
[3.x] Disable the frontpage view on new installs

### LINKED ISSUE #549: Disable to /node page
I don't think we should have this default Drupal page?
https://demo.localgovdrupal.org/node

It gets setup by default by Drupal.

andybroomfield:
Fix #549
for 3.x branch

Disable the /node view

### DIFF
--- a/localgov.install
+++ b/localgov.install
@@ -26,6 +26,13 @@ function localgov_install() {
   $frontpage_view = View::load('frontpage');
   $frontpage_view->setStatus(FALSE)->save();
 
+  // Set front page to /user.
+  // This is so there is a default accessible front page on first install.
+  \Drupal::configFactory()
+    ->getEditable('system.site')
+    ->set('page.front', '/user')
+    ->save();
+
 }
 
 /**
--- a/localgov.install
+++ b/localgov.install
@@ -6,6 +6,7 @@
  */
 
 use Drupal\user\Entity\User;
+use Drupal\views\Entity\View;
 
 /**
  * Implements hook_install().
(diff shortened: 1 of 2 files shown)


andybroomfield:
So the issue here is that
````
$this->drupalGet('<front>');
$this->assertSession()->statusCodeEquals(Response::HTTP_OK);
````
Is trying to load the front page, which as this PR disabled the /node view and no other front page exits, won't now have anything there.

I did add an additional test with 
````
// Test that the /node view is disabled.
$this->drupalGet('/node');
$this->assertSession()->statusCodeEquals(Response::HTTP_NOT_FOUND);
````

So I guess its down to how we want to proceed.
We could remove the test and accept that no default front page gets created, or we could install a dummy front page in place and test for it.

markconroy:
In the Annertech starterkit, we set /user as the frontpage, so we don't have the /node view, but it also gets us around the issue of not having any content yet.

ekes:
All in favour of disabling `/node`, but we should set `<front>` directly then, probably to some welcome to you new site default content?
I'm almost surprised we didn't already, but maybe that was just with all the time spent on onboarding ux for microsites.

andybroomfield:
Have changed the default to /user for new installs.
I noticed an odd behaviour with this change when installing localgov_demo which sets a subsite as the default home page.
It requires a cache clear for the new front page to take effect, otherwise the user login page still shows. I assume this will be the case with setting other custom home pages too (though they take place in the context of site building).

finnlewis (approved):
Good for now, thanks. @ekes suggested we might also want to add a piece of default content in the future.

PR #613, opened 2023-08-24
[2.x] Disable the frontpage view on new installs

### LINKED ISSUE #549: Disable to /node page

andybroomfield:
Fix #549
for 2.x branch

Disable the /node view


ekes:
Discussion on #612

finnlewis (approved):
Good for now, thanks. @ekes suggested we might also want to add a piece of default content in the future.

PR #632, opened 2023-10-16
Feature/3.x/631 php82

finnlewis:


finnlewis:
@millnut here's a pull request to try to help with https://github.com/localgovdrupal/localgov_project/pull/131

millnut:
@finnlewis looks good to me, it appears we install from the 3.x-dev on the workflow so this will need to be merged to 3.x in order for the patch to apply based on this line https://github.com/localgovdrupal/localgov/actions/runs/6533159900/job/17740888455?pr=632#step:10:3

millnut (approved):

PR #634, opened 2023-10-23
[3.x] Update gin_login to 2.x

### LINKED ISSUE #633: Update to Gin Login 2
```
composer why drupal/gin_login
localgovdrupal/localgov 3.0.0-beta1 requires drupal/gin_login (^1.0@RC) 
```
But this is no longer the recommended or supported version.

andybroomfield:
Fix #633

Gin login 1.x no longer supported.


stephen-cox (approved):
The single test failure is from Geo Field Map which also needs to be updated to a supported version.

PR #642, opened 2023-11-09
Patch entity_browser to address PHP 8.2 test failures.

finnlewis:
Trying to fix the failing test in https://github.com/localgovdrupal/localgov/actions/runs/6801391950/job/18492295931?pr=636

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "drupal/gin": "^3.0@alpha",
         "drupal/gin_login": "^1.0@RC",
         "drupal/gin_toolbar": "^1.0@beta",
+        "drupal/entity_browser": "^2.9",
         "drupal/preview_link": "^2.1@alpha",
         "drupal/simple_sitemap": "^4.1",
         "drupal/search_api": "^1.21",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,6 +17,7 @@
         "drupal/gin": "^3.0@alpha",
         "drupal/gin_login": "^2.0.3",
         "drupal/gin_toolbar": "^1.0@beta",
+        "drupal/entity_browser": "^2.9",
```
andybroomfield, line 20: Do we need this here?

andybroomfield (approved):
Otherwise tests now passing so this is good to go!

composer.json
finnlewis: Yes, it seems so! I think it only patches if the require is in the same composer.json

I'm still not 100% sure though.

Here's a test run without the require which has failures: 
https://github.com/localgovdrupal/localgov/actions/runs/6810487415

Here's the one with the require, which passed: 
https://github.com/localgovdrupal/localgov/actions/runs/6810600296

PR #645, opened 2023-11-19
Set PHP 8.1 as minimum PHP version

### LINKED ISSUE #463: PHP version support
Actually - we should probably follow Drupal's recommendations:

https://www.drupal.org/docs/system-requirements/php-requirements#versions

PHP version | Supported by Drupal | First Drupal release with support | Recommended?
-- | -- | -- | --
7.3 | 9.x | 8.6.4 (note 1 below) | No
7.4 | 9.x | 8.8.3 | No
8.0 | 9.x | 9.1.0 | Yes
8.1 (note 2 below) | 9.x, 10.x | 9.3.0 | Yes

So support but not recommend PHP 7.3 and 7.4

_Originally posted by @finnlewis in https://github.com/localgovdrupal/localgov/issues/424#issuecomment-1093120221_

andybroomfield:
Fix #463 

Specify minimum PHP version in composer.json like Drupal core.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "localgovdrupal/localgov_subsites": "^2.3.0",
         "localgovdrupal/localgov_scarfolk": "^1.1.2",
         "localgovdrupal/localgov_workflows": "^1.2.0",
-        "php": ">=8.1.0",
+        "php": ">=8.1.0"
     },
     "require-dev": {
         "drupal/core-dev": "^10.0",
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Looks good to me!

andybroomfield:
So the question here is, as we depend on Drupal core that will always set PHP 8.1 as minimum anyway, but this would give us a chance to set a minimum here if we wanted to set PHP 8.2 in future.

PR #654, opened 2023-12-13
Entity browser has been fixed upstream.

siliconmeadow:
Entity browser 8.x-2.10 has incorporated this patch now.


MatthewWilkes:
@siliconmeadow Potentially silly question here (I am not a PHP person, after all!), but should we up the version pin for entity_browser to the fixed version also?

stephen-cox (approved):
Looks fine. Thanks @siliconmeadow

stephen-cox:
> Potentially silly question here (I am not a PHP person, after all!), but should we up the version pin for entity_browser to the fixed version also?

No need to update the version requirement as the given restraint (`"drupal/entity_browser": "^2.9"`) will install all 2.* releases of the module.

PR #678, opened 2024-02-19
Update to latest version of the unpublised content access patch

### LINKED ISSUE #670: The reference unpublish content patch is failing on Drupal 10.1
This is the patch added in #336 and updated in #667 to allow editors to reference unpublished content.

On latest update I'm now seeing 
> Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2024-01-07/2845144-78.patch

Updating to the latest patch appears to resolve, though I'm going to have to test it's doing the right think as the diff is a lot shorter.

https://www.drupal.org/project/drupal/issues/2845144#comment-15389555

andybroomfield:
Fix #670

Uses the patch in comment 87.
This is the evolution of our current patch, rather than the shortned version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
                 "Fix PHP 8.2 deprecation notices": "https://www.drupal.org/files/issues/2023-05-22/3343964-5.patch"
             },
             "drupal/core": {
-                "Users can't reference unpublished content even when they have access to it. See https://www.drupal.org/project/drupal/issues/2845144": "https://www.drupal.org/files/issues/2024-01-07/2845144-78.patch"
+                "Users can't reference unpublished content even when they have access to it. See https://www.drupal.org/project/drupal/issues/2845144": "https://www.drupal.org/files/issues/2024-02-13/2845144-87.patch"
             }
         }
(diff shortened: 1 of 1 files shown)


stephen-cox (commented):
This fixes the patch failure which makes me think we should get it sooner than later, but I get different test failures on the update test, so perhaps someone else can double check @finnlewis @ekes @Adnan-cds?

finnlewis:
The new patch applies, but when running phpunit tests locally, I get: 

```
There was 1 error:

1) Drupal\Tests\localgov\Functional\LocalGovUpdateTest::testUpdate
PHPUnit\Framework\Exception: Segmentation fault (core dumped)
```

Edit: I also get the same error without the patch.

finnlewis:
The patch works, I have created a separate issue for the error I get: https://github.com/localgovdrupal/localgov/issues/680

finnlewis (approved):
I think we should merge this, make a release, then fix the failing tests which look to be unrelated.

millnut (approved):

Issue #600, opened 2023-08-04
Previewing more than a single page at a time


willguv:
Preview links only work for single pages. This means editors have to send multiple links to service colleagues for them to check guides, step by step, directories, whole content sections... it's not ideal

In Cumbria we got around this issue using separate environments, and Entity Share to push content between them https://cumbria.localgov.blog/2022/05/31/using-entity-share-with-localgov-drupal/

This may not be practical for many councils. It would be good to explore options

willguv:
Big issue for Essex - also inconsistent about what's shown. A big issue for stakeholders who're being asked a lot of "an email with 20 links isn't helping"

Also more urgent than other issues as content reviewing is happening all the time

Haringey has a single read only log in to workaround this

Newcastle uses Entity Share

Content designers can share a variety of content types from single guides to whole sections, so solution needs to work for all scenarios

andybroomfield:
Is also an issue at BHCC, we have used the login with limited role in the past.
Does entity share require a seperate provisoned environment just for testing?
That might be impractical. Would it be possible to extend preview link or create a new module that grants access to a limited set of pages? Maybe setting as a cookie instead of long links??

willguv:
Hi Andy, yes Entity Share requires a separate environment - agree that this isn't practical for a lot of councils. It was very useful for us in Cumbria though

I'd like us to look at extending preview link functionality or creating something new

Can't do Product Group today sadly (you should have a cancellation) but am catching up with Stephen soon to talk about this and a few other issues. Can invite you?

willguv:
Wondering if we could add another workflow state "Accuracy check" which can only be viewed by a user with limited permissions

Then create a single user account with this role, and details can be shared with services

Thinking aloud here...

willguv:
Easiest - provide a "Service contact" read only login, as Haringey has done. Users would be able to view all unpublished content, but we don't think this is too big an issue. Users would have to find their content again

Also talk at Tech Group about options @finnlewis @andybroomfield @stephen-cox 

---
Other options

Workflow access - a way of restricting view of unpublished content (eg by url or entry in services menu) . Provide documentation on config options

Workspace - could be created for staging etc or for a single guide which could be shared with a service contact

https://www.drupal.org/project/workspace

willguv:
Andy: 

Brighton takes read only login approach, and also sometimes publishes on staging environment. Looking at Entity Share

BHCC comms like the temporary link idea, but needs to work for multiple pages

Thoughts:

[...]

PR #692, opened 2024-03-28
Added LocalGov Preview Link module

stephen-cox:
This PR adds a new module that makes it easy for content editors to add all pages in a guide or step by step to a preview link.

The module adds a new button that will auto-populate the entities on a preview link page.

image

Part of #600

### DIFF
--- a/modules/localgov_preview_link/README.md
+++ b/modules/localgov_preview_link/README.md
@@ -1,0 +1,3 @@
+# LocalGov Drupal Preview Link
+
+Enhancements to Preview Link functionality for LocalGov Drupal.
--- a/modules/localgov_preview_link/localgov_preview_link.info.yml
+++ b/modules/localgov_preview_link/localgov_preview_link.info.yml
@@ -1,0 +1,7 @@
+name: 'LocalGov Drupal Preview Link'
+type: module
+description: 'LocalGov Drupal enhancements for the Preview Link module.'
+package: LocalGov Drupal
+core_version_requirement: ^10 || ^11
+dependencies:
+  - preview_link:preview_link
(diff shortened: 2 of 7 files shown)


stephen-cox:
@Adnan-cds @andybroomfield @ekes @finnlewis @and anyone else.

I'm not sure this is the best way to add the achieve this, so your thoughts are welcome.

What we want to achieve here is an easy way to add all guides, step by steps and subsite pages to the preview link entities list.

The approach I've taken is to put all the code to find the entities into a new module in the profile. It would be cleaner and possibly more extensible if each module supplied it's own logic to say whether adding multiple entities like this is supported and, if so, supply the logic to get all the related entities.

I suspect the Drupal way to do this is through a plugin system, but this seems overkill to me. But maybe not?

Thoughts?

andybroomfield:
Nice, thanks @stephen-cox.
I'd condider making this a seperate module and not limiting it to the profile, as it's likley to have utility in other cases and not all LGD sites install via the localgov install profile. (Profile modules can't AFAIK be used outside the profile).

finnlewis:
Discussing in Merge Tuesday meeting with @ekes : 

It would be good to include this somewhere other than the profile. 

The main arguments for not putting it in the profile: 

 - some people don't use the profile
 - profiles will probably be deprecated soon in favour of recipes
 - profiles can now be uninstalled https://www.drupal.org/node/3432357

So maybe we should move this sub-module to localgov_core (if we assume everyone wants it), or have it as a separate project (if it is useful as a standalone project).

finnlewis:
Discussing at Merge Tuesday with @stephen-cox : 

 - Relatively straight forward to add this to the https://www.drupal.org/project/preview_link module as a plugin and service
 - We can then implement the plugins in our modules.
 - Estimating around 10 hours to get it all done

I think that sounds like a great direction of travel!

stephen-cox:
Closing this as there's now an MR adding this functionality to the Preview Link module: https://git.drupalcode.org/project/preview_link/-/merge_requests/28

Issue #709, opened 2024-05-17
Content Lock module (v2.4) breaks the site


AWearring:
Content Lock module (v2.4) breaks the site with the following error: `Cannot redeclare Drupal\content_lock\ContentLock\ContentLock::$time`
Looks like we need to remove the patch for content_lock in https://github.com/localgovdrupal/localgov/blob/3.x/composer.json

See Slack conversation: https://localgovdrupal.slack.com/archives/CSE44GLSE/p1715953228601509

PR #710, opened 2024-05-17
Remove content_lock patch

AWearring:
Removing the content_lock patch as this is included in v2.4 of content_lock and revised the requirement for content_lock to ^2.4

See issue https://github.com/localgovdrupal/localgov/issues/709

And Slack conversation: https://localgovdrupal.slack.com/archives/CSE44GLSE/p1715953228601509

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "drupal/admin_toolbar": "^3.0",
-        "drupal/content_lock": "^2.3",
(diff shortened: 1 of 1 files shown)


millnut (approved):
Looks good to me @AWearring, and based on our slack it is indeed caused by the content_lock module now having a release which fixes issues with PHP 8.2 and the patch we include causes `$time` to be redeclared.

This also fixes the error that we get on the project workflow @finnlewis for the same reason above.

PR #715, opened 2024-05-21
Added views block for revision log

anthonylindsay:
uses core views to create a view of node revision log entries and dates, rendered as a block.
Uses a contextual filter to filter to a node ID.
Designed for display on a node.

### DIFF
--- a/config/install/views.view.revision_log.yml
+++ b/config/install/views.view.revision_log.yml
@@ -1,0 +1,278 @@
+uuid: f20313f2-0a70-4df0-ba43-ba517491abb5
+langcode: en-gb
+status: true
+dependencies:
+  module:
+    - node
+    - user
+id: revision_log
+label: 'Revision Log'
+module: views
+description: ''
+tag: ''
+base_table: node_field_revision
+base_field: vid
+display:
+  default:
+    id: default
+    display_title: Default
+    display_plugin: default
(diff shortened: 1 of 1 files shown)


config/install/views.view.revision_log.yml
```diff
@@ -0,0 +1,278 @@
+uuid: f20313f2-0a70-4df0-ba43-ba517491abb5
```
markconroy, line 1: Can we remove the UUID from this please

markconroy (changes requested):
This looks very handy.

I think it might be worth doing 2 more things

1. Use "HTML List" as the formatting option instead of "Unformatted list", and
2. Link the revision date to the revisions tab for that node.

Here's a screenshot of what I mean (I can't add it to this PR since it's outside of the LGD group).

Screenshot 2024-05-22 at 11 42 03

===

finnlewis:
Discussed at Merge Tuesday, we'll hang tight for the changes requested.

markconroy:
I've created a PR for this with the requested changes and some other updates as well. It's here: https://github.com/localgovdrupal/localgov/pull/735

I think we can close this one now.

finnlewis:
I think we're closing this in favour of https://github.com/localgovdrupal/localgov/pull/735

PR #721, opened 2024-05-27
adds/enables msqrole module

### LINKED ISSUE #674: Install 'Masquerade role' and create a dedicated role for it
https://www.drupal.org/project/msqrole

Andy said at a PG that Brighton is using this successfully - covers the bases of features being visible and useable by specific roles

Masquerade was previously suggested https://www.drupal.org/project/masquerade but it's not covered by the Drupal security policy

markconroy:
Closes #674 

This PR is to add and enable msqrole module only. It _does not_ add a dedicated role for the purpose. 

If we want a dedicated role, let's create a new (generic) role for these type things.

===


ekes:
Saw the discussion about this. And I didn't get that this would need to be installed on prod. Maybe only on stage and dev sites?

markconroy:
@ekes I think this is probably worth chatting about at Merge Tuesday and/or Tech Drop-in.

andybroomfield:
Leaving for the disuccion in #674

andybroomfield:
Just updated to fix merge conflict pending discussion.
It would be good to get either MSQrole or masquerade installed.
From BHCC viewpoint we also install msqrole on prod to help diagnose issues with roles.
We have a mild preference for msrole as 
- It has security support
- Lets us test role coverage without having to find existing user
- Has an overlay to indicate using masquraded roles and easy to remove them (the masqureade block conflicts with Gin theme during tests).

Pros for masqureade
- Lets you take over an actual user account, so if someone says they are having a problem, you view the site as if you are logged in as them.

localgov.install
```diff
@@ -80,9 +87,25 @@ function localgov_update_9502() {
 }
 
 /**
- * Enable msqrole module.
+ * Update existing sites to use entity_usage for media.
  */
```
ekes, line 91: I'm a bit confused as to what's happened here?

This is still the MR for enabling some masquerade system?

andybroomfield: @ekes I think it was just the merge confict as the Update existing sites to use entity_usage for media PR went in first and they both originally used the same update hook number 9503.

andybroomfield: Actual change (move this to update 9504:  https://github.com/localgovdrupal/localgov/blob/33c491c3e366d34ed6ae0ac0201b808c863b6deb/localgov.install#L89-L112

markconroy:
I'm going to close this PR. We agreed to go with masquerade module at the product group meeting last week, now that masquerade has security team coverage and is the preferred choice from Drupal Starshot initiative.

PR #723, opened 2024-05-27
adds 'copy-to-clipboard' button for preview_link

### LINKED ISSUE #722: Quick copy button for preview links
Hi. 

Is it possible to have a quick copy button for preview links? Would save having to click and highlight the link and Ctrl C.

Something similar to YouTube or the 2 squares copy icon?

Thanks.

PreviewCopy

markconroy:
Closes #722 

Here's a PR for #722, while we are waiting on the patch sent to Drupal.org to be committed.

===

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,7 @@
                 "Users can't reference unpublished content even when they have access to it. See https://www.drupal.org/project/drupal/issues/2845144": "https://www.drupal.org/files/issues/2024-02-13/2845144-87.patch"
             },
             "drupal/preview_link": {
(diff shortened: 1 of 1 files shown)


stephen-cox:
This looks to work well. Although I have a couple of suggestions:

* The patch file contains sections for all the commits, it would be easier to read if these were all squashed like they are in https://git.drupalcode.org/project/preview_link/-/merge_requests/29.diff
* Very minor - maybe the button should be added between the link and the expiry date rather than after the expiry date, as it applies to the link.

stephen-cox (approved):
New patch looks good

PR #729, opened 2024-06-11
Adds simple_media_bulk_upload

### LINKED ISSUE #726: Add 'simple media bulk upload' to core
Some councils have installed https://www.drupal.org/project/simple_media_bulk_upload to speed up loading of multiple media files

@benhillsjones has used it in Haringey and rates it, was installed by Annertech

Let's install it on Prod and try it out

markconroy:
Closes #726 

===


composer.json
```diff
@@ -4,25 +4,40 @@
...
+            "package": {
+                "name": "enyo/dropzone",
+                "version": "6.0.0-beta.2",
+                "type": "drupal-library",
+                "dist": {
+                    "url": "https://github.com/dropzone/dropzone/releases/download/v6.0.0-beta.2/dist.zip",
+                    "type": "zip"
+                }
+            }
+        },
```
andybroomfield: Not so sure about this, is this fixing the version that will get installed? It means that it will need to be updated each time. Maybe we need to look at the [composer merge plugin](https://packagist.org/packages/wikimedia/composer-merge-plugin) that some modules like [webform](https://git.drupalcode.org/project/webform/-/blob/6.2.x/composer.libraries.json?ref_type=heads)  or maybe use [asset packagist](https://asset-packagist.org/) instead?

markconroy: It's fixing the version, yes, but there's hasn't been a new version in about 2 years so we probably won't be under too much pressure.

markconroy: Hi Andy, any further thoughts on this, or are we okay to approve it?

finnlewis:
Stimulating a wider discussion on how to manage libraries like this.

 - composer merge plugin
 - composer asset packagist (https://asset-packagist.org/) 
 - specific repositories like above.

@andybroomfield thinks that most are moving to composer merge.

finnlewis:
@Adnan-cds mentioned : 

> While considering the Dropzonejs module, please consider this: https://www.drupal.org/project/dropzonejs/issues/3316184

finnlewis:
@andybroomfield is up for taking this forward: 

 - this might need to all be in the localgov_project

andybroomfield:
Drafting the composer merge plugin method here https://github.com/localgovdrupal/localgov_project/pull/173

willguv:
@finnlewis @markconroy let's try to push this up tomorrow - it's reliant a beta library that's 4 years old and I feel like this is good enough to try

finnlewis:
Discussing in Merge Tuesday: 

@stephen-cox leaning towards composer-merge plugin rather than the npm asset packagist

@Adnan-cds is suggesting the approach here, but changing it a little to avoid specifying the dist and version, rather targeting the git repo and allow us to require a more open constraint. 

Note: we really want to align on a preferred solution for LocalGov Drupal in general.

@stephen-cox points out that if we do require the libraries in the project composer.json, end users can decide whether to override this approach or not. So for us, we define our libraries in the localgov_project composer.json and remove the specific library definition here.

finnlewis:
So it looks like @andybroomfield added the libraries to the localgov_project composer.json.

https://github.com/localgovdrupal/localgov_project/commit/1e749dd73bcf887d765c3b1bfe47b0ce86447db8

So can we just remove the specification of the repo from here? :

https://github.com/localgovdrupal/localgov/pull/729/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R7-R19

andybroomfield:
Noting that I get the same issue as https://github.com/localgovdrupal/localgov_assets/issues/2 when installing this PR (enyo/dropzone will install under vendor/enyo/dropzone)

composer.json
stephen-cox: As per my comments on https://github.com/localgovdrupal/localgov/issues/726#issuecomment-2659353202, I think we remove this. Merge https://github.com/localgovdrupal/localgov_project/pull/173 and include in the release notes instructions for the Composer Merge Plugin and Asset Packagist.

composer.json
```diff
@@ -4,25 +4,41 @@
...
         "drupal/core": "^10.0",
+        "drupal/dropzonejs": "^2.10",
         "drupal/gin": "^3.0 || ^4.0",
         "drupal/gin_login": "^2.0.3",
         "drupal/gin_toolbar": "^1.0 || ^2.0",
         "drupal/entity_browser": "^2.9",
         "drupal/masquerade": "^2.0",
         "drupal/preview_link": "^2.1@alpha",
+        "drupal/simple_media_bulk_upload": "^1.0",
+        "drupal/simple_sitemap": "^4.1",
```
stephen-cox: This line is not relevant to this PR.

composer.json
```diff
@@ -4,25 +4,41 @@
...
         "drupal/redirect": "^1.10",
         "drupal/search_api": "^1.21",
         "drupal/simple_sitemap": "^4.1",
         "drush/drush": ">=10",
+        "enyo/dropzone": "^6.0.0-beta.2",
```
stephen-cox: This line isn't needed for the Composer Merge Plugin.

localgov.info.yml
```diff
@@ -35,12 +35,14 @@ install:
   - admin_toolbar:admin_toolbar
   - admin_toolbar:admin_toolbar_tools
   - entity_browser:entity_browser
+  - dropzonejs:dropzonejs
```
stephen-cox: This line isn't needed as simple_media_bulk_upload already has a dependency on this module.

stephen-cox (changes requested):
I think this can radically simplified when using the Composer Merge Plugin or Asset Packagist.

To add another conversation starter, should this be moved to the localgov_media module (part of Core)?

markconroy:
This is ready for testing again.

stephen-cox (approved):
Works for me (I adjusted the order of packages slightly)

PR #735, opened 2024-06-26
Feature/734/adds revisions log view

### LINKED ISSUE #734: Show published/ last updated dates on pages
PaulB raised this at Product Group and in Slack

> Is there anything on the board about adding published/last updated dates to webpages? Is this feature already available?

Stephen:

> It's easy enough to add this sort of functionality; just need to adjust the display options
> One thing we can do is make sure they display nicely when enabled

### To do

- Discuss need at content group
- Decide what needs to be shown and where on the page
- Decide whether dates are shown on every page or selected ones (Paul suggests selected ones)

markconroy:
Closes #734 

Updates the PR from Anthony Lindsay with the some changes
- Sets the view to only show 5 items
- Sets a mini pager if there are more than 5 items
- Uses Ajax so there is no page refresh when using the pager
- Renamed the view to `Change log` instead of `Revision log`
- Sets the view to use the `localgov_` namespace to avoid conflicts with any custom views that might exist
- Sets the view to use a HTML list instead of unformatted list

This PR will only work for new installs. Do we want an update hook to also add it to existing sites? It won't affect existing sites as it will only create the view, not place the block - devs/councils will need to do that themselves.

Screenshot 2024-06-26 at 17 20 17

===


finnlewis:
Do we perhaps want to move this out of the profile with a view towards recipes etc ?

ekes:
> Do we perhaps want to move this out of the profile with a view towards recipes etc ?

It's just config? So easily moved into a recipe when we have one to put it into.

finnlewis:
Discussing at Merge Tuesday, @markconroy reminded me that this potentially publishes revision log notes that were made a long time ago, before the decision to enable this. 

So we'll close this in favour of adding a more custom approach, possibly a custom revision message field?

PR #736, opened 2024-06-28
fix: config type breaking tests

millnut:
Once https://github.com/localgovdrupal/drupal-container/pull/33 and the other `drupal-container` PRs are merged and a new docker image is created for each PHP version, this PR will fix the error below which is reported in tests across ~21 modules.

```
1) Drupal\Tests\localgov_content_lock\Functional\ContentLockTest::testContentLockConfiguration
Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for system.cron with the following errors: system.cron:logging variable type is integer but applied schema class is Drupal\Core\TypedData\Plugin\DataType\BooleanData

/app/web/core/lib/Drupal/Core/Config/Development/ConfigSchemaChecker.php:98
/app/web/core/lib/Drupal/Component/EventDispatcher/ContainerAwareEventDispatcher.php:111
/app/web/core/lib/Drupal/Core/Config/Config.php:230
/app/web/core/lib/Drupal/Core/Config/ConfigInstaller.php:396
/app/web/core/lib/Drupal/Core/Config/ConfigInstaller.php:149
/app/web/core/lib/Drupal/Core/ProxyClass/Config/ConfigInstaller.php:75
/app/web/core/lib/Drupal/Core/Extension/ModuleInstaller.php:326
/app/web/core/lib/Drupal/Core/ProxyClass/Extension/ModuleInstaller.php:83
/app/web/core/includes/install.inc:567
/app/web/core/includes/install.core.inc:1125
/app/web/core/includes/install.core.inc:695
/app/web/core/includes/install.core.inc:572
/app/web/core/includes/install.core.inc:122
/app/web/core/lib/Drupal/Core/Test/FunctionalTestSetupTrait.php:315
/app/web/core/tests/Drupal/Tests/BrowserTestBase.php:570
/app/web/core/tests/Drupal/Tests/BrowserTestBase.php:370
/app/web/profiles/contrib/localgov/modules/localgov_content_lock/tests/src/Functional/ContentLockTest.php:33
/app/vendor/phpunit/phpunit/src/Framework/TestResult.php:729
```

PR #737, opened 2024-07-01
fix: change uninstall string check to boolean check

millnut:
This fixes the following error:

```
1) Drupal\Tests\localgov\Functional\LocalGovProfileTest::testLocalGovDrupalProfile
Failed asserting that 'The following reasons prevent the modules from being uninstalled: The 'LocalGov Drupal' install profile requires 'LocalGov Core'; The install profile 'LocalGov Drupal' is providing the following module(s): localgov_login_redirect' contains "module is required".

/var/www/html/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:122
/var/www/html/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:56
/var/www/html/web/profiles/contrib/localgov/tests/src/Functional/LocalGovProfileTest.php:36
/var/www/html/vendor/phpunit/phpunit/src/Framework/TestResult.php:729
```

Prior to Drupal 10.3 the assert check was checking the error message contains `module is required`, with Drupal 10.3 the error message does not contain this string and is much longer.

```
The following reasons prevent the modules from being uninstalled: The 'LocalGov Drupal' install profile requires 'LocalGov Core'; The install profile 'LocalGov Drupal' is providing the following module(s): localgov_login_redirect
```

This PR changes the check to use a boolean and just checking if an error message exists.

The failure in the workflow that paratest hides the true error message, is fixed in a previous PR https://github.com/localgovdrupal/localgov/pull/736

### DIFF
--- a/tests/src/Functional/LocalGovProfileTest.php
+++ b/tests/src/Functional/LocalGovProfileTest.php
@@ -33,7 +33,8 @@ class LocalGovProfileTest extends BrowserTestBase {
       $this->fail('Uninstalled localgov_core module.');
     }
     catch (ModuleUninstallValidatorException $e) {
-      $this->assertStringContainsString('module is required', $e->getMessage());
+      $hasError = strlen($e->getMessage()) > 0 ? TRUE : FALSE;
+      $this->assertTrue($hasError);
     }
 
     // Test localgov_core:localgov_roles submodule is enabled.


ekes:
Just hit re-run all jobs as there was a paratest - ie no idea what - phpunit fail. I'm assuming we're not expecting a fail on this?

millnut:
Yep paratest hides the actual failures, when running phpunit one failure is https://github.com/localgovdrupal/localgov/issues/738 which is still failing and the other failure is fixed in this PR https://github.com/localgovdrupal/localgov/pull/736

PR #751, opened 2024-07-31
Adds new install step, which calls a new hook.

rupertj:
## What does this change?

Adds a new step to the installer, which calls a new hook (hook_localgov_post_install) to let other modules perform tasks.

## How to test

Use the version of localgov_core from this MR: https://github.com/localgovdrupal/localgov_core/pull/212. That will implement the hook in this MR and use it to set up blocks in the installer.

### DIFF
--- a/localgov.api.php
+++ b/localgov.api.php
@@ -1,0 +1,16 @@
+<?php
+
+/**
+ * Run a task during the site installation process.
+ *
+ * This is intended for work that needs to happen when installing a localgov
+ * site, that can't happen in a module's hook_install(). This hook is invoked
+ * later in the install process, when everything bar the importing of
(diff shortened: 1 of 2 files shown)


rupertj:
See the instructions in this MR for how to test this: https://github.com/localgovdrupal/localgov_core/pull/212

stephen-cox (approved):
This looks to me to be a nice solution to the problems found in localgovdrupal/localgov_core#212

Would be nice to have @ekes input as well, but happy to approve.

ekes (approved):
Looks fine to me, just reminds me we really _must_ find the time to start working on moving to recipes already. Because this sort of thing I think we'll have to do differently.

PR #755, opened 2024-08-11
Compatibility with Workspaces

ekes:
## What does this change?

Workspaces has issues with core, pathauto, and redirect that are experienced with LocalGov Drupal.

### Core

https://www.drupal.org/project/drupal/issues/3179199

Most commonly the automatically created pathauto revision id will clash with a previous revision id of some content that is being published, causing (because of not checking entity type with revision id) an erroneous error that there is content in  a non-default workflow state in the space. Very confusing.

#### How to test

Automated tests on the patch. Manual steps (I'd advise the last one) to reproduce on the issue.

### Pathauto

https://www.drupal.org/project/pathauto/issues/3283769

Main problem experienced in manual testing with LGD was when something with a path pattern dependent on a parent (like services parent) was in a space that was published the paths were regenerated and the parent path would be /node/123 at the point of publishing, so the child ended up with node/123/child-title. With the patch paths are generated in the space, and not regenerated when the space is published.

#### How to test

Tediously create content in spaces with paths that depend on a parent till it goes wrong.

### Redirect

https://www.drupal.org/project/redirect/issues/3431260

This one allows redirect to be somewhat workspaces aware. Notes on my earlier thoughts https://www.drupal.org/project/redirect/issues/3114103#comment-15714027 

#### How to test (and problem)

Create content in a space, then change its title. It will generate a redirect, but this will fail with a fatal error as it tries to create a redirect.

Can use workspaces with LGD.

Issues don't change data structure, so should be fine to apply. If they stop applying or a different approach is taken we had Workspaces working till then and will have to move to any other fix.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,11 +67,11 @@
                 "Add a 'copy to clipboard' feature for preview_link": "https://www.drupal.org/files/issues/2024-05-28/3449121-9.patch"
             },
             "drupal/pathauto": {
-                "Allow path generation inside of a workspace - and importantly don't regenerate when publishing space": "https://www.drupal.org/files/issues/2024-04-08/3283769-10.patch"
+                "Allow path generation inside of a workspace - and importantly don't regenerate when publishing space https://www.drupal.org/project/pathauto/issues/3283769": "https://www.drupal.org/files/issues/2024-04-08/3283769-10.patch"
             },
             "drupal/redirect": {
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Yeah baby! This does seem to fix the problems with Workspaces where changing url aliases in the stage workspace generated fatal errors.

Nice work @ekes !

PR #756, opened 2024-08-12
Update patch for Redirect 1.10+

ekes:
See https://www.drupal.org/project/redirect/issues/3057250#comment-15723568

I've not _yet_ _*tested*_ the patch though. Just that it applies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "drupal/gin_toolbar": "^1.0@beta",
         "drupal/entity_browser": "^2.9",
         "drupal/preview_link": "^2.1@alpha",
-        "drupal/redirect": "^1.6",
+        "drupal/redirect": "^1.10",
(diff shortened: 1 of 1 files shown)


AWearring:
I've tested this on my local build and it resolves the issue that is present without the patch

ekes:
Looking at the patch it's still moving the code from RedirectSourceWidget::formElement into RedirectForm. There were changes to the ::formElement part, which is validation (hence the conflict), and they aren't moved over to RedirectForm. But at a scan I think they're UI related not actual validation. So it should be fine.

finnlewis:
Tested and confirmed without the patch I can replicate the behaviour on https://www.drupal.org/project/redirect/issues/3057250

Patch applies and the behaviour is fixed.

Thanks @ekes !

finnlewis (approved):

PR #759, opened 2024-08-15
updates patch for preview link

markconroy:
## What does this change?

Preview Link has had a new release, so I had to re-roll the patch that is against it for "Copy to clipboard". This updates the patch that we can now use.

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
             },
             "drupal/preview_link": {
                 "Automatically populating multiple preview link entities #3439968": "https://www.drupal.org/files/issues/2024-05-22/3439968-4.diff",
-                "Add a 'copy to clipboard' feature for preview_link": "https://www.drupal.org/files/issues/2024-05-28/3449121-9.patch"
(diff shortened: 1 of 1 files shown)


AWearring:
Hmm, doesn't work for me on my local build
In Firefox I get an `Uncaught TypeError: navigator.clipboard is undefined` error
In Chrome I get `Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')`

AWearring:
> Hmm, doesn't work for me on my local build In Firefox I get an `Uncaught TypeError: navigator.clipboard is undefined` error In Chrome I get `Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')`

It appears that this could be an issue with my local virtual PC - having reinstalled the previous version of preview link (and patch) I get the same errors - this version works on our staging and live sites

finnlewis:
Patch applies for me and the preview link functionality appears to work as expected.

finnlewis (approved):

PR #777, opened 2024-09-30
adds/enables masquerade module

### LINKED ISSUE #674: Install 'Masquerade role' and create a dedicated role for it

markconroy:
Closes #674 

## What does this change?

Adds/enables masquerade module.

---

### DIFF
--- a/localgov.profile
+++ b/localgov.profile
@@ -45,12 +45,3 @@ function localgov_post_install_task(): void {
     $hook();
   });
 }
-
-/**
- * Enable masquerade module.
- */
-function localgov_update_10001() {
-  if (!\Drupal::service('module_handler')->moduleExists('masquerade')) {
-    \Drupal::service('module_installer')->install(['masquerade']);
-  }
-}
--- a/localgov.profile
+++ b/localgov.profile
@@ -45,3 +45,12 @@ function localgov_post_install_task(): void {
     $hook();
   });
 }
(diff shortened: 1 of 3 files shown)


localgov.profile
```diff
@@ -45,3 +45,12 @@ function localgov_post_install_task(): void {
     $hook();
   });
 }
+
+/**
+ * Enable masquerade module.
+ */
+function localgov_update_10001() {
+  if (!\Drupal::service('module_handler')->moduleExists('masquerade')) {
+    \Drupal::service('module_installer')->install(['masquerade']);
```
finnlewis: Hi @markconroy thanks for this.

I'm not sure we want to enable masquerade for all existing sites. 

If I'm using https://www.drupal.org/project/msqrole (which we are on a number of sites) then I don't want this as well. 

I suggest we remove the update hook and let people know in the release notes that it is now available if desired.

Do we also want to add permissions for this to the default install?

finnlewis (changes requested):

finnlewis (approved):
Thanks @markconroy !

PR #780, opened 2024-10-10
Update README.md to add DDEV docs link

tanc:
## What does this change?

Adds a link to the README.md to the DDEV docs. As a developer this README is the first thing I saw and I quickly did a search for DDEV to see how to set things up. There was a line about DDEV but no guidance, so I fumbled around a bit until I saw the general link to the docs. Within the docs there is an excellent page on setting up with DDEV so I suggest linking to it from the README.


MariosORION:
I had that in mind when I was rewriting these docs a while ago but never added this - good to have a link in there! 👍

finnlewis (approved):
Thanks @tanc !

Issue #771, opened 2024-09-17
Preview links not rewriting


willguv:
Flagged by Jamie

> @stephen-cox I've created a preview link for a guide that has 11 pages. I clicked 'Add all pages to this guide' and [when I view this preview link logged in](https://www.wirral.gov.uk/preview-link/node/11905/fe929db5-5c1b-4689-aa24-91909d3e5109) I can see all the pages. When I'm logged out I can only see the first page. Does it only work for users who are logged in? Most colleagues I use this with will not have an account on the website. 

 _Originally posted by @Dixonj1 in [#600](https://github.com/localgovdrupal/localgov/issues/600#issuecomment-2349209664)_

willguv:
@stephen-cox I think it's urgent we track down this problem and patch/ suggest a fix to maintainers. This feature working properly is more important to content editors than character counts. Can you let me kow how we go about this please? Thanks

andybroomfield:
Just confirming the issue applies to a fresh copy of localgov drupal with localgov_demo enabled.

## Steps to reproduce:
1. Create an unpublished guide overview page
2. Create several unpublished guide pages, link them to the guide overview.
3. On the guide overview page, click preview link
4. On the preview link page, click and all pages for this guide
5. Copy the preview link and open in a private browser window (anon user)

## Expected result

Page shows with links to all guide pages in the top navigation box.

## Actual result

Anon user can only preview the top level page, navigaton box does not show additional links.
(Though if they paste the links in it does give them access, so it's not a permission issue).

<img width="1045" alt="Screenshot 2024-09-25 at 1 43 56 PM" src="https://github.com/user-attachments/assets/f530e86e-24e8-4574-84ab-c1a71537ff26">


<img width="665" alt="Screenshot 2024-09-25 at 1 43 42 PM" src="https://github.com/user-attachments/assets/5e043a71-3831-4930-8a87-29ef15a5d953">

And pasting in the link to the second page directly.
<img width="859" alt="Screenshot 2024-09-25 at 1 48 21 PM" src="https://github.com/user-attachments/assets/ccc2d2a2-063a-46bc-bc44-f16934deb854">

willguv:
Thanks for testing this @andybroomfield !

andybroomfield:
[...]

PR #785, opened 2024-10-18
Patch Preview Link for bug where references to other entities being previewed don't display

stephen-cox:
## What does this change?

Patched Preview Link for issue #771, drupal.org issue: https://www.drupal.org/project/preview_link/issues/3481523

The patch changes the node access hook so Preview Links access handler is run for all pages included in a preview link entity. Without the hook it only runs for the page being viewed.

## How to test

See https://github.com/localgovdrupal/localgov/issues/771#issuecomment-2373991933 on how to reproduce the issue this should fix.

Previewing more than one page becomes easier and works as people expect.

This changes an access handler so has the potential to allow more than it should.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,8 @@
             },
             "drupal/preview_link": {
                 "Automatically populating multiple preview link entities #3439968": "https://www.drupal.org/files/issues/2024-05-22/3439968-4.diff",
-                "Add a 'copy to clipboard' feature for preview_link": "https://www.drupal.org/files/issues/2024-08-15/3449121-10.patch"
+                "Add a 'copy to clipboard' feature for preview_link": "https://www.drupal.org/files/issues/2024-08-15/3449121-10.patch",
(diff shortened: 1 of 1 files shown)


stephen-cox:
Tests failures are fixed with https://github.com/localgovdrupal/localgov/pull/787

ekes:
To clarify if I understand what your patch is doing:
Previously the access check override looked at the route, checked if it is in preview mode, checked if the route entity is one that can be previewed.
Now the access check, looks if it's in preview mode, runs through all entities that can be previewed, and sees if there's a match?

ekes:
Hit wrong button, was just the question :-)

stephen-cox:
> Previously the access check override looked at the route, checked if it is in preview mode, checked if the route entity is one that can be previewed.
> Now the access check, looks if it's in preview mode, runs through all entities that can be previewed, and sees if there's a match?

That's right - previously is only checked access for the entity being previewed but now checks whether the entity is included in a related preview link.

finnlewis:
@ekes is happy with this, let's merge it!

PR #787, opened 2024-10-18
Fix failing update test 

stephen-cox:
## What does this change?

Looks like there was a small text change in the step by step wording that was causing a test failure. This updates the check for this change.

## How to test

Ensure tests pass!


ekes (approved):
As the test passes, and it's looking for a sane string in a restricted region, seems good.

finnlewis (approved):

PR #794, opened 2024-12-11
Remove code to add non-existent administrator role.

ekes:
We have added a localgov_admin role that is supplied by localgov_admin_role https://github.com/localgovdrupal/localgov_core/pull/234 and included in this profile https://github.com/localgovdrupal/localgov/commit/0d19a7cfa92fc3dbe6b4d7db7aff174292231a17 This role is added to user 1 by core (and drush) automatically as (the only) administrator role at the time of install.

Credit to @NikLP for spotting that there was a role on User 1 at time of install that was not actually a role on the site.

### DIFF
--- a/localgov.install
+++ b/localgov.install
@@ -5,7 +5,6 @@
  * Install functions for the LocalGovInstall installation profile.
  */
 
-use Drupal\user\Entity\User;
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Looks good to me!

PR #800, opened 2024-12-18
Sets form to no-validate, so we can use Drupal validation instead

### LINKED ISSUE #503: Disable HTML5 Validation and use Drupal's validation instead.
This is a follow up to https://github.com/localgovdrupal/localgov_base/issues/367

https://www.drupal.org/project/disable_html5_validation

markconroy:
Closes #503 

## What does this change?

HTML 5 validation is not very good. Drupal's form validation is much better.

This simply removes HTML 5 validation from forms - same approach as [Disable HTML5 Validation](https://www.drupal.org/project/disable_html5_validation) module - so Drupal's form validation can take over instead.

Before:

Screenshot 2024-12-18 at 14 07 16

After: 

Screenshot 2024-12-18 at 14 06 15

### DIFF
--- a/localgov.profile
+++ b/localgov.profile
@@ -5,6 +5,8 @@
  * Enables modules and site configuration for a Localgov site installation.
  */
 
+use Drupal\Core\Form\FormStateInterface;
+
 /**
  * Implements hook_page_attachments().
  */
@@ -44,4 +46,11 @@ function localgov_post_install_task(): void {
   \Drupal::moduleHandler()->invokeAllWith('localgov_post_install', function (callable $hook, string $module) {
     $hook();
   });
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function localgov_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $form['#attributes']['novalidate'] = 'novalidate';
 }
--- a/localgov.profile
+++ b/localgov.profile
@@ -51,6 +51,6 @@ function localgov_post_install_task(): void {
 /**
  * Implements hook_form_alter().
  */
-function localgov_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+function localgov_form_alter(&$form, FormStateInterface $form_state, $form_id): void {
   $form['#attributes']['novalidate'] = 'novalidate';
 }


andybroomfield:
Would we be able to put this behind a setting / config option? Be useful to introduce gradually to test. 
Could be some issues if custom modules are dependant on html5 validation.

andybroomfield:
Also I think this might be better in localgov_core?

andybroomfield:
I tested and it works well, in that I get the Drupal validation errors and it makes a lot more sense. I'll need to check BHCC code to see if we are dependant on html5 validation anywhere.

markconroy:
I wonder would it make sense as a theme setting? Or do we want it in a module to disable HTML5 validation everywhere?

finnlewis:
> I tested and it works well, in that I get the Drupal validation errors and it makes a lot more sense. I'll need to check BHCC code to see if we are dependant on html5 validation anywhere.

@andybroomfield [in the distant past, you commented](See: https://github.com/localgovdrupal/localgov_base/issues/367#issuecomment-1422541027
) that: 

>Although this is focused on user login, and we should still fix the issue anyway for these cases, we should move away from using HTML5 validation if we want to match GDS guidelines.
see https://design-system.service.gov.uk/patterns/validation/

I see that the GDS guidelines specifically say "Turn off HTML5 validation." 

So I agree with your previous comment and think we should merge this and roll out better validation by default. 

What's the argument for making it optional?

If it really needs to be optional, we can use https://www.drupal.org/project/disable_html5_validation and install by default, allowing people to uninstall if desired.

finnlewis (approved):
I think we should go with this and happy to approve, but  @andybroomfield I'd like you to be OK with this too.

finnlewis:
Sorry, only just spotted the PHP static analysis failure, so just added a commit to address that.

finnlewis:
All test pass now too :)

finnlewis:
@millnut Quick question: are hooks (like this one) still OK in the .profile file or do we want to move them to a module?

millnut:
I agree with @andybroomfield this should probably live in `localgov_core` so we don't have to move it at a later date with recipes.

markconroy:
I have created the same PR in localgov_core now so we can merge that one instead if we are in agreement.

https://github.com/localgovdrupal/localgov_core/pull/267

finnlewis:
Thanks @markconroy 

Closing this one in favour of https://github.com/localgovdrupal/localgov_core/pull/267

PR #807, opened 2025-01-14
Removes localgov_content_lock and localgov_login_redirect from the profile.

### LINKED ISSUE #778: Pull submodules localgov_content_lock and localgov_login_redirect out of the profile
In preparation for recipes we need to pull localgov_content_lock  and localgov_login_redirect from the profile into their own composer packages so that we can include them in the base recipe

rupertj:
Fix #778


finnlewis (approved):
Tested locally and it just works! Nice one @rupertj !

As an aside: I note that we do not enable the content_lock functionality by default, I wonder if we might want to do this? (in a another issue).

ekes:
Tested with a clean current install, with php caching on for the autoloader, and updated:

`drush cr` after update works with this one :sweat_smile: 

But it does require a cache rebuild after update.

finnlewis:
Thanks @ekes ! I'll merge and tag a release.

PR #808, opened 2025-01-17
feat: update gin to recommended version

millnut:
## What does this change?

Update gin to versions recommended by the maintainer


composer.json
```diff
@@ -14,9 +14,9 @@
         "drupal/admin_toolbar": "^3.0",
         "drupal/content_lock": "^2.4",
         "drupal/core": "^10.0",
-        "drupal/gin": "^3.0@alpha",
+        "drupal/gin": "^4.0",
         "drupal/gin_login": "^2.0.3",
-        "drupal/gin_toolbar": "^1.0@beta",
+        "drupal/gin_toolbar": "^2.0",
```
andybroomfield: Maybe
```
"drupal/gin": "^3.0 || ^4.0",
"drupal/gin_login": "^2.0.3",
"drupal/gin_toolbar": "^1.0 || ^2.0",
```
As some themes might have breaking changes in the major version bump.
(Gin Tweaks type modules using the old variable names is one that we encountered).

millnut: Thanks @andybroomfield I'll add that in. I believe the only changes are dropping D9 support but I'll update just incase

finnlewis (approved):
Tested locally and installs and works nicely. Thanks @millnut :)

Issue #811, opened 2025-01-22
Remove dependency on search_api


rupertj:
The profile declares a dependency on search_api, but as far as I can see, doesn't actually depend on search_api directly at all.

PR #812, opened 2025-01-22
Remove dependency on search_api.

rupertj:
Fix for #811.


ekes:
Was added here https://github.com/localgovdrupal/localgov/commit/1db9e6fe2b49a476295fb7988b70a66a98451c00 when we were patching it here too. That's gone. So the dependency can go to. It'll be required by the appropriate modules.

ekes (approved):

PR #820, opened 2025-02-04
Disable HTML5 form validation by default #503

ekes:
Discussion:
https://github.com/localgovdrupal/localgov_core/pull/267#issuecomment-2618854939 and before.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   tests:
-    uses: localgovdrupal/localgov_shared_workflows/.github/workflows/test-module.yml@1.x
+    uses: localgovdrupal/localgov_shared_workflows/.github/workflows/test-module.yml@1.x-d10
     with:
       project: 'localgovdrupal/localgov'
       project_path: 'web/profiles/contrib/localgov'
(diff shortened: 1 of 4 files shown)


ekes:
Am I correct that it's failing to install [disable_html5_validation](https://www.drupal.org/project/disable_html5_validation) because it requires D10, but states `Drupal: >=9` Is this the automagically created composer.json file from Drupal.org maybe? (It doesn't seem to have one)

millnut:
Hi @ekes I believe it's because `localgov` doesn't yet have D11 support so it's failing the D11 install because of [this line](https://github.com/localgovdrupal/localgov/blob/3.x/composer.json#L15), and as the branch was created from 3.x it only has D10 support currently

ekes:
@millnut So we won't be able to get the tests to pass for D10?

finnlewis:
Passing to @millnut  to try to get the tests running for D10 at least!

millnut:
@finnlewis @ekes  I've temporarily disabled the `needs` check for the phpunit step so it will continue even if the build fails so we can see the phpunit tests for D10.

This error/issue will go away once the other modules have D11 support, and we can then declare the profile as working with D11

stephen-cox:
I've updated the workflow so it only tests against D10 for now. We'll need to revert this once the profile can declare D11 support.

ekes (commented):
Dropping testing against 11.x till we want to test it seems fine to me.

MR itself is simple enough and agreed.

Ah! And I can't approve this because I created the original MR!

finnlewis (approved):
Thanks folks, let's get this in,

PR #829, opened 2025-03-04
Set Preview Link expiry in days rather than seconds

### LINKED ISSUE #124: Set Preview Link expiry in days
Right now we're asking content designers to do this

Image

/admin/config/content/preview_link

Can we set this to days please?

The field title should be "Link expiry"

The helper text should be "The number of days before a preview link expires."

## Update docs

Once done, Will needs to update https://docs.localgovdrupal.org/content/features/workflow.html#previews

stephen-cox:
Fixes  localgovdrupal/localgov_workflows#124

## What does this change?

Patches the Preview Link module so the expiry time for preview links is set by days rather than seconds.

## How to test

Go to /admin/config/content/preview_link and change the preview link expiry in days.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,8 @@
             "drupal/preview_link": {
                 "Automatically populating multiple preview link entities #3439968": "https://www.drupal.org/files/issues/2024-05-22/3439968-4.diff",
                 "Add a 'copy to clipboard' feature for preview_link": "https://www.drupal.org/files/issues/2024-08-15/3449121-10.patch",
-                "References to other entities being previewed don't display #3481523": "https://www.drupal.org/files/issues/2024-10-18/3481523-4.patch"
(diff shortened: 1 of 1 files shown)


ekes:
Not checked why yet, but

```
  - Installing drupal/preview_link (2.2.0-alpha2): Extracting archive
  - Applying patches for drupal/preview_link
    https://www.drupal.org/files/issues/2025-03-04/3510967-3.patch (Set Preview Link expiry in days #3510967)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2025-03-04/3510967-3.patch
```

ekes:
But using https://git.drupalcode.org/project/preview_link/-/merge_requests/36.diff directly it does work. So when the patch is updated to be equal to that it's good.

ekes (approved):
Working patch now applies.

PR #837, opened 2025-03-12
3.2.0-alpha2 release

finnlewis:
To fix fatal error on status report.


finnlewis:
@stephen-cox thinks we might be able to disable the check for VERSION in our shared workflow.

PR #852, opened 2025-04-29
DO NOT MERGE: Adding for testing upgrade to field_group module.

ekes:
Added for running tests with upgraded version. Assuming I got the aliases correct.


ekes:
Working just PHP8.4 as ever.

PR #871, opened 2025-06-28
feat: allow new gin theme and gin_toolbar module

millnut:
## What does this change?

Add new gin theme and gin_toolbar that support Drupal 11.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,9 +13,9 @@
     "require": {
         "drupal/admin_toolbar": "^3.0",
         "drupal/core": "^10.0 || ~11.1.0",
-        "drupal/gin": "^3.0 || ^4.0",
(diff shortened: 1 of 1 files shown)


markconroy:
@millnut There's a merge conflict in this branch from some recent changes.

millnut:
Thanks for the flag @markconroy I'll take a look at that shortly

millnut:
@markconroy that should be fixed now

millnut:
@tonypaulbarker would be good to get this merged so we don't get an unsupported release message for gin in Drupal 11.2

markconroy (approved):
LGTM

tonypaulbarker:
@stephen-cox just to note the new versions of gin and gin toolbar require Drupal 11.2 and above so I don't think we need make this change below 3.2.x

PR #872, opened 2025-07-09
Remove conjunctions from URLs

### LINKED ISSUE #817: Write doc to remove conjunctions from URLs
Hi @dedavidson I'm looking at a few D&G improvements which we agreed would be pushed back to LGD at some point

We did some work on url conjunctions - is this easily translatable into a PR for new sites please?

dedavidson:
## What does this change?

This adds additional words to the pathauto default config that are to be ignored when creating a URL

Coming from #817

### DIFF
--- a/config/install/pathauto.settings.yml
+++ b/config/install/pathauto.settings.yml
@@ -1,0 +1,20 @@
+enabled_entity_types:
+  - user
+punctuation:
+  hyphen: 1
+verbose : FALSE
+separator : '-'
+max_length : 100
+max_component_length: 100
+transliterate : TRUE
+reduce_ascii : FALSE
+case : TRUE
+ignore_words: 'a, an, and, are, as, at, before, but, by, for, from, get, if, is, in, into, it, like, of, off, on, onto, or, per, since, so, than, the, this, that, to, up, via, when, where, with, you'
+update_action : 2
+safe_tokens:
+ - alias
+ - path
+ - join-path
+ - login-url
+ - url
+ - url-brief
--- a/config/install/pathauto.settings.yml
+++ b/config/install/pathauto.settings.yml
@@ -12,9 +12,9 @@ case : TRUE
 ignore_words: 'a, an, and, are, as, at, before, but, by, for, from, get, if, is, in, into, it, like, of, off, on, onto, or, per, since, so, than, the, this, that, to, up, via, when, where, with, you'
 update_action : 2
 safe_tokens:
- - alias
- - path
- - join-path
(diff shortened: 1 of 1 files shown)


markconroy:
@dedavidson I'm going to remove myself as a reviewer of this PR. 

I'm not going to block it, but I think we should keep _all_ words in our URL and not remove any of them, for predictability's sake. 

For example a node with Title "It is up to you" will end up with a url of `/it-is-up-to-you` but a node with title "It is up to you David" will get a URL of `/david`. And if we already had a page at `/david` then the "It is up to you David" page will have a URL of `/david-0`. I don't like that level of unpredictability (and how do we let content designers know what words are available and what are not).

willguv:
Thanks for your comment @markconroy 

This originated at Dumfries & Galloway and the need was to follow GDS style link structure - is this right @dedavidson?

Copying @finnlewis @tonypaulbarker as this could be a Tech Governance discussion

markconroy:
Here's some guidance from GOVUK

> Dashes should be used to separate words within URLs to ensure they are easy to read.   For example, https://www.gov.uk/running-a-limited-company.  (see 5, 7 and 15 for exceptions on campaign landing pages and service sub-domain URLs.)

> Articles (a, an, the) and other superfluous words should not be used.  For example, /benefits or /benefits-guides rather than /a-guide-to-benefits

I love the irony that the first thing to remove `a, an, the` are actually left in in the first example they give `https://www.gov.uk/running-a-limited-company`

They also say URL patterns are different for landing pages vs internal pages, and external sites (specifically mentioning Local Authorities) may use different patterns to GOVUK.

willguv:
Cheers for linking to the guidance @markconroy 

Some councils will want this, and it might become more important as GDS starts to work with local government

How about including it in the profile but switched off, with the option for councils to switch on if they like

finnlewis:
It is very easy for someone to change the behaviour at any point after installation by changing the configuration in pathauto.

See https://demo.localgovdrupal.org/admin/config/search/path/settings

image


So the question remains, what is the best practice and what do we want to ship with at install?

@tonypaulbarker @ekes @andybroomfield @Adnan-cds @stephen-cox I'll add it to the agenda for tech governance and we can discuss there if needed. 

Any further opinions welcome here too!

andybroomfield:
Could we message out to the content group to see what the general best practice is and follow that.
I'd just be reluctant to update it for existing sites who may have their own rules.
Chances are many don't, and so a way of posting an advisory so it can be updated, or a reset to recommended?

PR #885, opened 2025-08-25
Show last user to update content

### LINKED ISSUE #856: Add "Last updated by" filter
@AWearring mentioned they have a filter for 'Last updated by' which is not necessarily the author, but the person who last edited it. This could be a nice addition too!

_Originally posted by @finnlewis in https://github.com/localgovdrupal/localgov/issues/855#issuecomment-2876163480_

markconroy:
Closes #856 

## What does this change?

- Removes 'author' from content list
- Adds 'last updated by' instead
- Adds exposed filter for 'last updated by' (but this users user id instead of user name, so needs some work)

---

### DIFF
--- a/config/install/views.view.content.yml
+++ b/config/install/views.view.content.yml
@@ -792,7 +792,7 @@ display:
           field: revision_uid
           relationship: none
           group_type: group
-          admin_label: 'Last updated'
+          admin_label: 'Last updated by'
           entity_type: node
           entity_field: revision_uid
           plugin_id: standard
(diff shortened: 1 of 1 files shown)


finnlewis:
Notes from Merge Tuesday: 
 - this is not ready to be merged
 - the user id wants to be displayed as a user name

finnlewis:
@markconroy I have this working locally now, with just core views config changes :)

image

Just pushed the changes, wanna give it a spin? 

I got the tip off from here: https://drupal.stackexchange.com/questions/314561/views-exposed-filter-cant-input-username-to-filter-only-works-with-user-id

:)

markconroy:
Works a treat. Approved.

PR #888, opened 2025-09-01
Adds editoria11y module

### LINKED ISSUE #859: Consider adding Editoria11y module
I think [editoria11y](https://www.drupal.org/project/editoria11y) might be more feature rich than our sa11y module.

I especially like the reports page it creates to let us see what pages have a11y issues.

Image

This is part of Drupal CMS so seems it's promoted as a best practice.

markconroy:
Closes #859 

## What does this change?

Adds "LocalGov Editoria11y" module, which pulls in the Editoria11y module and configures the permissions for it.

## How to test

1. Install the module
2. Edit content as "LocalGov Editor" role
3. Each page should have the Editoria11y widget in the bottom right corner
4. There should be a general reports page here: https://localgov.ddev.site/admin/reports/editoria11y

More accessible LGD sites!!!

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "drupal/gin_login": "^2.0.3",
         "drupal/gin_toolbar": "^1.0 || ^2.0",
         "drupal/entity_browser": "^2.9",
+        "drupal/localgov_editoria11y": "^1.0.0@alpha",
         "drupal/disable_html5_validation": "^2.0",
         "drupal/localgov_utilities": "^1.0@beta",
         "drupal/masquerade": "^2.0",


finnlewis:
Very cool, thanks @markconroy , it seems to work as advertised!

It is not installed by default. 

When enabled: 

A user with the 'editor' (localgov_editor) role  has the Editoria11y widget in the bottom right corner

Some questions: 

1. Why does the install need to remove permissions before adding them in the module file? https://git.drupalcode.org/project/localgov_editoria11y/-/blob/1.0.x/localgov_editoria11y.install?ref_type=heads#L41
2. Should the namespace be drupal/ rather than localgovdrupal/ https://git.drupalcode.org/project/localgov_editoria11y/-/blob/1.0.x/composer.json?ref_type=heads#L2
3. Are we happy including it as an alpha release or do we want to make at least beta before inclusion?

finnlewis:
@markconroy this look OK? https://git.drupalcode.org/project/localgov_editoria11y/-/merge_requests/1/diffs

markconroy:
> Why does the install need to remove permissions before adding them in the module file? https://git.drupalcode.org/project/localgov_editoria11y/-/blob/1.0.x/localgov_editoria11y.install?ref_type=heads#L41

If not, all roles will get the editorially widget, but I took the position that initially only localgov_editor should get it, rather than everyone such as emergency_publisher. So I'm just removing the permission from others and then adding it to editors.

Thinking about it, we should probably remove permissions from _all_ roles initially, and then add it back in for editors and admins only, rather than hardcoding the list of roles it should be removed from.

Issue: https://www.drupal.org/project/localgov_editoria11y/issues/3544375
MR: https://git.drupalcode.org/project/localgov_editoria11y/-/merge_requests/3

@finnlewis can you test that ^

> Should the namespace be drupal/ rather than localgovdrupal/ https://git.drupalcode.org/project/localgov_editoria11y/-/blob/1.0.x/composer.json?ref_type=heads#L2

Yes, I've merged your PR for that.

> Are we happy including it as an alpha release or do we want to make at least beta before inclusion?

Let's get to beta first.

finnlewis:
@tonypaulbarker points out that this has more features than sa11y, so we probably want to phase sa11y out.

finnlewis:
I will test this again, thanks @markconroy :)

AWearring:
For localgov_sa11y we added a config form so that we could allow devs to exclude parts of the site (such as header/footer etc) as editors won't be able to change anything in those areas. We also added options for positioning - our preference would be to have it top-right, others will want it elsewhere: https://github.com/localgovdrupal/localgov_sa11y/blob/1.x/src/Form/Sa11yOptionsForm.php

Can we do something similar for localgov_editora11y?

tonypaulbarker:
@AWearring what would be cool then would be to look at adding controls to turn things on and off according to the context where they should be fixed. Some things can probably be fixed upstream. Some might be for microsites controllers. In another case, a web manager is doing an exercise and wants to raise a ticket for developers to fix an issue so they should have the option to see everything. It sounds complicated but I can see the value in narrowing down the notices.

finnlewis:
Discussing in Merge Tuesday, @tonypaulbarker still keen to make sure we have feature parity with sa11y insofar as we can define the regions of the page that we want to review.

finnlewis:
Created this issue for feature parity https://www.drupal.org/project/localgov_editoria11y/issues/3548191

tonypaulbarker:
Discussing with @msayoung 

We think that:
sa11y has contrast checking feature that editoria11y lacks
sa11y has a setting to move its widget

A possible route forward is sa11y on the left, editoria11y on the right. 

We think that editoria11y could be installed by default on new sites (as long as it is turn-off-able with sa11y becoming optional

tonypaulbarker (approved):

Issue #891, opened 2025-09-08
Update redirect patch for redirect 1.12


tonypaulbarker:
New 1.12 compatible version of the code for this issue is available 

https://www.drupal.org/project/redirect/issues/3057250

PR #893, opened 2025-09-09
Update redirect Validation issue patch and require 1.12 and above for 3.2.x

tonypaulbarker:
Closes https://github.com/localgovdrupal/localgov/issues/891 for 3.2.x


tonypaulbarker:
For this 3.2.x MR the tests that are failing should be addressed after https://github.com/localgovdrupal/localgov/pull/882 is merged

Issue #895, opened 2025-09-11
Error on fresh install


rupertj:
We just need to check this array index exists before using it.

<img width="1266" height="1195" alt="Image" src="https://github.com/user-attachments/assets/88d5556c-bcc2-4e8c-8ba3-71eea9f34ebf" />

rupertj:
Actually, this isn't just a case of checking the array exists. Our hook in the profile needs to run after the system module hook that provides the data we change, and for some reason on my install it's not. Changing the order of the hooks so ours is consistently after system fixes this.

andybroomfield:
I'm not encountering this, on a fresh install though both `drush si localgov -y` and though the UI installer, deleting the database between each.

<img width="1417" height="1056" alt="Image" src="https://github.com/user-attachments/assets/8de23497-1419-4fbc-ba12-e1adb8ddd58e" />

This is on Drupal 10.5.3

Any other modules being enabled along with the profile? Something that would clear out the page attachments array?

rupertj:
The site's on platform.sh, and was installed via the UI. I don't think there's any other modules that could be causing this error, as I was copying the setup from another site that doesn't have this problem, but this is what it has on it:

```
  "require": {
    "composer/installers": "^2.0",
    "drupal/ai_provider_aws_bedrock": "^1.1@beta",
    "drupal/core-composer-scaffold": "^10.0",
    "drupal/core-project-message": "^10.0",
    "drupal/core-recommended": "^10.0",
    "drupal/redis": "^1.6",
    "drush/drush": "^12",
    "localgovdrupal/localgov": "^3.1",
    "localgovdrupal/localgov_publications_importer": "dev-feature/1.x/28-background-processing",
    "platformsh/config-reader": "^2.4"
  },
```

andybroomfield:
Running xDebug I inserted this code in the profile
```php
function localgov_module_implements_alter(&$implementations, $hook) {
  if ($hook == 'page_attachments') {
    // Move bhcc events plus to last hook.
    $implementated_modules = array_keys($implementations);
  }
}
```
and inspected which modules where implementing the page_attachements hook.

<img width="1096" height="422" alt="Image" src="https://github.com/user-attachments/assets/1f088c93-c412-486a-83a7-940133e15261" />

I see that the order is 'contextual, system, localgov'. 
Are you able to do a similar thing in the environment where the site installs and see what the order is.
[...]

PR #904, opened 2025-10-06
Use the appropriate hook 

lazysoundsystem:
Ref: #895

## What does this change?

As suggested in #895, uses hook_page_attachments_alter() instead of hook_page_attachments().

## How to test

View source.

Look for the line `<meta name="Generator" content="Drupal 10 (LocalGov Drupal | https://localgovdrupal.org)" />` (and no error message as reported on #895)

Just noting it was originally using the other hook because the Thunder profile did it that way: https://github.com/localgovdrupal/localgov_core/pull/177#issuecomment-1576805488  

## Images
N/A

## Accessibility
N/A

### DIFF
--- a/localgov.profile
+++ b/localgov.profile
@@ -7,16 +7,18 @@
 
 /**
  * Implements hook_page_attachments_alter().
+ *
+ * Use LocalGov Drupal for site generator metatag.
  */
 function localgov_page_attachments_alter(array &$attachments): void {
   foreach ($attachments['#attached']['html_head'] as &$html_head) {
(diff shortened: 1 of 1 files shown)


tonypaulbarker (approved):
Functional testing looks good. I didn't see the error reported at https://github.com/localgovdrupal/localgov/issues/895 but perhaps I didn't have all the necessary conditions for triggering that.

lazysoundsystem:
Thanks @tonypaulbarker I couldn't re-produce the error either, but looks like it's now in the right place.

PR #916, opened 2025-11-08
fix: bump geofield_map to supported version

millnut:
This bumps geofield_map dev dependency to a supported version to fix

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
     },
     "require-dev": {
         "drupal/core-dev": "^10.0",
-        "drupal/geofield_map": "^3.0"
+        "drupal/geofield_map": "^11.1"
     },
     "minimum-stability": "dev",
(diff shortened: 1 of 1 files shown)


finnlewis:
Do we still want to keep geofield_map?

We moved away from the geofield_map widget a while ago:

https://github.com/localgovdrupal/localgov_geo/wiki/Upgrading-localgov_geo-fom-1.x-to-2.x-%E2%80%90-manual-configuration-review

and

https://github.com/localgovdrupal/localgov_geo/releases/tag/2.0.1

This is still causing some issues recently, with both @AWearring  and @nccchris having problems with https://www.drupal.org/project/field_formatter_class/issues/3473891

So if we want to keep geofield_map for any reason, upgrading it makes sense.

If not, shall we remove the dependency?

